PR #22, opened 2012-09-10
[KNIFE-9] Include mingw\bin to the PATH when bootstrapping to build native gems

petecheslock:
https://tickets.opscode.com/browse/KNIFE-9

Tested this on my local - windows node bootstrapped and was able to install the "thin" gem during the first run,  without it failed.

### DIFF
--- a/lib/chef/knife/core/windows_bootstrap_context.rb
+++ b/lib/chef/knife/core/windows_bootstrap_context.rb
@@ -79,7 +79,7 @@ CONFIG
         end
 
         def start_chef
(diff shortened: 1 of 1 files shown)


btm:
This is stale, so I'm going to close it. The important question we were discussing from the ticket was:

```
One issue with this is that if you don't add the same PATH to your systems in your chef runs, any pattern that runs successfully here using chef_gem isn't going to work on subsequently chef runs and this is likely to be confusing. We think it would be more appropriate to consider adding this PATH as part of the MSI installation. However, would that affect the first Chef run via bootstrap? What's the more complete solution?
```

Issue #119 looks like a different bug, but is the same topic so that'll stay open for now.

PR #59, opened 2013-10-11
Put messages in ui, not in STDOUT

alex-slynko-wonga:
Copied part of print_data function from chef/knife/ssh


btm:
I think this was fixed by 941b2610 / #164.

PR #72, opened 2014-02-07
[KNIFE-449] knife-windows Does Not Respect --bootstrap-version Option

sean-horn:


btm:
Duplicate of #75? Was fixed by cad8f261 #131.

PR #93, opened 2014-08-11
Update windows_bootstrap_context.rb

schade:
During bootstrap the chef-client is not currently set to either run or activate the chef-client service. Adding this bit, 'ADDLOCAL="ChefClientFeature,ChefServiceFeature"' to the end of the installation command on line 152 activates these two Features for windows during install.


btm:
Does this just changes the default for the checkboxes?

btm:
We're better off probably having a bootstrap setting/argument for this. I'm closing this PR and we'll track that feature in https://github.com/chef/knife-windows/issues/202.

Issue #133, opened 2014-12-08
bootstrap's creation of client.rb should mimic chef gem's code


jeremiahsnapp:
The code that creates the client.rb config during bootstrap should manage the same options that chef gem's code does.

Now that Chef Client 12 is out and SSL verification is enabled by default the knife-windows' code should auto-transfer `trusted_certs_dir` and manage `verify_api_cert` and `ssl_verify_mode`.

https://github.com/opscode/knife-windows/blob/0.8.2/lib/chef/knife/core/windows_bootstrap_context.rb#L47-L80

https://github.com/opscode/chef/blob/12.0.0/lib/chef/knife/core/bootstrap_context.rb#L56-L116

carpnick:
So is this closed/completed or being left open until released?

adamedx:
@carpnick, we'll keep this open to track getting it released -- that way it's clear when someone encounters it that the issue is still active in the released gem.

NimishaS:
@carpnick , fix is merged in master branch but not released yet. We can close this issue once the fix is released.

pburkholder:
@adamedx Any blockers to releasing 0.8.3 or 0.8.3.rc for this?

adamedx:
No, working on an RC release today, would be great to confirm this resolves all the issues once the rc is out.

adamedx:
We've uncovered some RC release blockers, so we won't have the release today. You can still try out master @carpnick / @pburkholder , but one some systems the removal of the event-machine dependency is causing non-deterministic behavior which we need to address before we have a usable RC.

If we can't fix this soon enough in master, the option will be to backport the SSL changes to a stable branch and release a patch.

kwilczynski:
@adamedx @NimishaS do y'all know when this will be resolved and when a new release would be cut? We have a customer (actually, more than one) who is blocked by this, and would like to see this fixed before they go live in production.

adamedx:
@btm, time to pull trigger on 0.8.3 -- that will fix the issue for users with chef-client 11.

btm:
@jeremiahsnapp 0.8.4.rc.0 should contain this fix now. could you please test it?

btm:
the following options aren't fully implemented still
    --bootstrap-template
    --[no-]node-verify-api-cert
    --node-ssl-verify-mode

I've copied over --node-ssl-verify-mode and --node-verify-api-cert and released 0.8.4.r.c1.

[...]

PR #147, opened 2014-12-19
Windows Chef Client version should be consistent with version for non-Windows bootstrap

adamedx:


adamedx:
@btm, this doesn't seem to have been merged -- did it get pushed? If you can merge this soon we can release this and work around #133.

btm:
pushed 0-stable.

btm:
I've merged 0-stable back into master. There was one test failing on master and one additional on my merge branch. I'm going to track that as separate work so we don't continue diverging from Clogeny's work on master where there area few subtle differences that make the merge complicated with every commit.

0-stable has been deleted and we won't use it anymore, we will be shipping directly from master.

PR #149, opened 2014-12-19
Adding knife wsman test to validate WSMAN/WinRM availability

smurawski:
This adds a command "knife wsman test" to help verify connectivity to remote WinRM endpoints.  Currently this can be tested from Windows nodes with PowerShell, but it is more difficult to validate connectivity cross platform.  This command will validate that the remote port is reachable and listening for management requests similar to Test-Wsman in PowerShell. 

Targets can be specified manually (as per the knife winrm command) or use search to identify target nodes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,11 +47,18 @@ This subcommand operates in a manner similar to [knife ssh](https://docs.chef.io
 ### knife wsman test
 
 Connects to the remote WSMan/WinRM endpoint and verifies the remote node is listening.  This is the equivalent of running Test-Wsman from PowerShell.  Endpoints to test can be specified manually, or be driven by search and use many of the same connection options as knife winrm.
+To test a single node using the default WinRM port (5985)
 
     knife wsman test 192.168.1.10 -m
-or
+
+or to test a single node with SSL enabled on the default port (5986)
+
+    knife wsman test 192.168.1.10 -m --winrm-transport ssl
+
+or to test all windows nodes registered with your Chef Server organization
 
     knife wsman test platform:windows
+
 
 ### knife bootstrap windows winrm
 
@@ -61,7 +68,7 @@ This subcommand operates in a manner similar to [knife bootstrap](https://docs.c
 
     knife bootstrap windows winrm ec2-50-xx-xx-124.compute-1.amazonaws.com -r 'role[webserver],role[production]' -x Administrator -P 'super_secret_password'
 
-### Use SSL for WinRM communication 
+### Use SSL for WinRM communication
 
 By default, the `knife winrm` and `knife bootstrap windows winrm` subcommands use a plaintext transport,
 but they support an option `--winrm-transport` (or `-t`) with the argument
--- a/README.md
+++ b/README.md
@@ -44,6 +44,15 @@ Or force a chef run:
 
 This subcommand operates in a manner similar to [knife ssh](https://docs.chef.io/knife_ssh.html)...just leveraging the WinRM protocol for communication. It also include's `knife ssh`'s "[interactive session mode](https://docs.chef.io/knife_ssh.html#options)"
 
+### knife wsman test
+
+Connects to the remote WSMan/WinRM endpoint and verifies the remote node is listening.  This is the equivalent of running Test-Wsman from PowerShell.  Endpoints to test can be specified manually, or be driven by search and use many of the same connection options as knife winrm.
+
+    knife wsman test 192.168.1.10 -m
+or
+
+    knife wsman test platform:windows
+
 ### knife bootstrap windows winrm
 
 Performs a Chef Bootstrap (via the WinRM protocol) on the target node. The goal of the bootstrap is to get Chef installed on the target system so it can run Chef Client with a Chef Server. The main assumption is a baseline OS installation exists. It is primarily intended for Chef Client systems that talk to a Chef server.
--- a/knife-windows.gemspec
+++ b/knife-windows.gemspec
@@ -15,7 +15,8 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version	= ">= 1.9.1"
   s.add_dependency "winrm", "~> 1.3"
-  s.add_dependency "winrm-s", "~> 0.2"
+  s.add_dependency "winrm-s", "~> 0.3.0.dev.0"
+  s.add_dependency "nokogiri"
 
 
   s.add_development_dependency 'pry'
@@ -26,3 +27,4 @@ Gem::Specification.new do |s|
   s.executables   = `git ls-files -- bin/*`.split("\n").map{ |f| File.basename(f) }
   s.require_paths = ["lib"]
 end
+
--- a/lib/chef/knife/windows_helper.rb
+++ b/lib/chef/knife/windows_helper.rb
@@ -20,6 +20,7 @@ require 'chef/knife'
 require 'chef/knife/winrm'
 require 'chef/knife/bootstrap_windows_ssh'
 require 'chef/knife/bootstrap_windows_winrm'
+require 'chef/knife/wsman_test'
 
 class Chef
   class Knife
@@ -27,7 +28,8 @@ class Chef
 
       banner "#{BootstrapWindowsWinrm.banner}\n" +
               "#{BootstrapWindowsSsh.banner}\n" +
-              "#{Winrm.banner}"
+              "#{Winrm.banner}\n" +
+              "#{WsmanTest.banner}" 
     end
   end
 end
--- a/lib/chef/knife/winrm.rb
+++ b/lib/chef/knife/winrm.rb
@@ -17,332 +17,42 @@
 #
 
 require 'chef/knife'
-require 'chef/knife/winrm_base'
+require 'chef/knife/winrm_knife_base'
 require 'chef/knife/windows_cert_generate'
 require 'chef/knife/windows_cert_install'
 require 'chef/knife/windows_listener_create'
+require 'chef/knife/winrm_session'
 
 class Chef
   class Knife
-    class Winrm < Knife
+    class Winrm < Knife 
 
-      @@ssl_warning_given = false
-
-      class Session
-        attr_reader :host, :output, :error, :exit_code
-        def initialize(options)
-          @host = options[:host]
-          url = "#{options[:host]}:#{options[:port]}/wsman"
-          scheme = options[:transport] == :ssl ? 'https' : 'http'
-          endpoint = "#{scheme}://#{url}"
-          opts = Hash.new
-          opts = {:user => options[:user], :pass => options[:password], :basic_auth_only => options[:basic_auth_only], :disable_sspi => options[:disable_sspi], :no_ssl_peer_verification => options[:no_ssl_peer_verification]}
-
-          options[:transport] == :kerberos ? opts.merge!({:service => options[:service], :realm => options[:realm], :keytab => options[:keytab]}) : opts.merge!({:ca_trust_path => options[:ca_trust_path]})
-          Chef::Log.debug("WinRM::WinRMWebService options: #{opts}")
-          Chef::Log.debug("Endpoint: #{endpoint}")
-          Chef::Log.debug("Transport: #{options[:transport]}")
-          @winrm_session = WinRM::WinRMWebService.new(endpoint, options[:transport], opts)
-          @winrm_session.set_timeout(options[:operation_timeout]) if options[:operation_timeout]
-        end
-
-        def relay_command(command)
-          remote_id = @winrm_session.open_shell
-          command_id = @winrm_session.run_command(remote_id, command)
-          Chef::Log.debug("#{@host}[#{remote_id}] => :run_command[#{command}]")
-          session_result = get_output(remote_id, command_id)
-          @winrm_session.cleanup_command(remote_id, command_id)
-          Chef::Log.debug("#{@host}[#{remote_id}] => :command_cleanup[#{command}]")
-          @exit_code = session_result[:exitcode]
-          @winrm_session.close_shell(remote_id)
-          Chef::Log.debug("#{@host}[#{remote_id}] => :shell_close")
-        end
-
-        def get_output(remote_id, command_id)
-          @winrm_session.get_command_output(remote_id, command_id) do |out,error|
-            print_data(@host, out) if out
-            print_data(@host, error, :red) if error
-          end
-        end
-
-        def print_data(host, data, color = :cyan)
-          if data =~ /\n/
-            data.split(/\n/).each { |d| print_data(host, d, color) }
-          elsif !data.nil?
-            print Chef::Knife::Winrm.ui.color(host, color)
-            puts " #{data}"
-          end
-        end
-      end
-
-      include Chef::Knife::WinrmBase
+      include Chef::Knife::WinrmCommandSharedFunctions     
 
       deps do
         require 'readline'
         require 'chef/search/query'
-        require 'winrm'
       end
 
       attr_writer :password
 
-      banner "knife winrm QUERY COMMAND (options)"
-
-      option :attribute,
-        :short => "-a ATTR",
-        :long => "--attribute ATTR",
-        :description => "The attribute to use for opening the connection - default is fqdn",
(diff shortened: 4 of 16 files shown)


btm:
@smurawski it'd be great if you could finish this up so we can include it in the upcoming 1.0.0 release with all the new, potentially confusing, winrm bits.

smurawski:
@btm Will do.  I'll get it fixed up pronto.

smurawski:
@btm @jdmundrawala @ksubrama Would love some other eyes on this.  I'm going to add some more tests around the new feature, but there is a pretty significant refactor which exposed a lot of problems with our existing tests and failures to resolve default parameters. (The change to the default of negotiate should have broke a lot of tests.)

lib/chef/knife/windows_helper.rb
```diff
@@ -27,7 +28,8 @@ class WindowsHelper < Knife
 
       banner "#{BootstrapWindowsWinrm.banner}\n" +
```
jdmundrawala: I think it's time to refactor this
```ruby
[BootstrapWindowsWinrm.banner, 
 BootstrapWindowsSsh.banner, 
 Winrm.banner, 
 WsmanTest.banner].join("\n")
```

smurawski: It can probably just go away, since it was more a rollup of "all" the knife-windows options under a Windows header.  It doesn't have the cert commands listed at all either.

smurawski: I take that back, it does have the other cert commands since they have the knife windows pattern.

lib/chef/knife/winrm.rb
```diff
@@ -17,100 +17,76 @@
...
-        @winrm_sessions ||= []
-        @winrm_sessions.push(session)
+        configure_session
+        validate_password
+        execute_remote_command        
       end
 
+      def execute_remote_command
+        begin
+          @longest = 0
```
jdmundrawala: i don't think this gets used anywhere

lib/chef/knife/winrm.rb
```diff
@@ -330,77 +163,36 @@ def check_for_errors!
...
-  knife[:winrm_ssl_verify_mode] = :verify_peer
-```
-* * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * *
-WARN
-        end
-      end
+      def success_return_codes
+        #Redundant if the CLI options parsing occurs
+        return [0] unless config[:returns]
+        return config[:returns].split(',').collect {|item| item.to_i}
```
jdmundrawala: the result of this function can probably be memoized

lib/chef/knife/winrm_knife_base.rb
```diff
@@ -0,0 +1,208 @@
...
+      def self.included(includer)
+        includer.class_eval do
+
+          @@ssl_warning_given = false
+
+          include Chef::Knife::WinrmBase
+          include Chef::Knife::WinrmSharedOptions          
+
+          def configure_session   
+            #configure_chef
```
jdmundrawala: commented out code

lib/chef/knife/winrm_knife_base.rb
```diff
@@ -0,0 +1,208 @@
...
+          include Chef::Knife::WinrmSharedOptions          
+
+          def configure_session   
+            #configure_chef  
+            resolve_session_options
+            resolve_target_nodes       
+            session_from_list(@list)
+          end
+
+          def resolve_target_nodes
```
jdmundrawala: The formatting in this function is really weird

lib/chef/knife/winrm_session.rb
```diff
@@ -0,0 +1,72 @@
...
+require 'winrm'
+
+class Chef
+  class Knife
+    class WinrmSession
+        attr_reader :host, :endpoint, :port, :output, :error, :exit_code
+        def initialize(options)
+          @host = options[:host]
+          @port = options[:port]
+          url = "#{options[:host]}:#{options[:port]}/wsman"
```
jdmundrawala: I think you can just use `host` and `port` now

lib/chef/knife/winrm_session.rb
```diff
@@ -0,0 +1,72 @@
...
+          scheme = options[:transport] == :ssl ? 'https' : 'http'
+          @endpoint = "#{scheme}://#{url}"
+          opts = Hash.new
```
jdmundrawala: don't need this, there is a new hash being created on the next line

jdmundrawala:
:+1: Awesome stuff @smurawski

lib/chef/knife/winrm_knife_base.rb
smurawski: Yeah, missed removing that.. Thanks.

lib/chef/knife/winrm_session.rb
smurawski: Cool.  Most of this is just pulled from what was before.

lib/chef/knife/winrm_knife_base.rb
smurawski: If you mean that one still takes a parameter (which is an instance variable), that is due to an expectation in the tests.  I should clean that test up so I don't have to pass the argument.

lib/chef/knife/winrm_session.rb
smurawski: In the case of assignment, won't that just create local variables?

lib/chef/knife/winrm_knife_base.rb
jdmundrawala: I was talking about the spacing

`when` does not match up with `case`, some blocks have 1 space instead of 2

lib/chef/knife/winrm_base.rb
```diff
@@ -114,7 +114,9 @@ def self.included(includer)
 
       def locate_config_value(key)
         key = key.to_sym
-        config[key] || Chef::Config[:knife][key]
+        value = config[key] || Chef::Config[:knife][key] || default_config[key]
```
btm: This entire method makes me nervous, or makes my scars from previous battles with argument precedence tingle.

Can't we just trust config[key] because it gets set similarly in Chef::Knife#merge_configs? https://github.com/chef/chef/blob/master/lib/chef/knife.rb#L341

lib/chef/knife/winrm_knife_base.rb
```diff
@@ -0,0 +1,207 @@
+#
+# Author:: Steven Murawski (<smurawski@chef.io)
+# Copyright:: Copyright (c) 2015 Opscode, Inc.
```
btm: Chef Software, Inc.

lib/chef/knife/winrm_knife_base.rb
```diff
@@ -0,0 +1,207 @@
...
+          include Chef::Knife::WinrmBase
+          include Chef::Knife::WinrmSharedOptions          
+
+          def configure_session               
+            resolve_session_options
+            resolve_target_nodes       
+            session_from_list
+          end
+
+          def resolve_target_nodes
```
btm: this looks like a replacement for Chef::Knife#configure_session. Could you open up an issue against Chef and leave a comment here that says we're duplicating that code until the bug is fixed, at which point we can use that method?

spec/unit/knife/wsman_test_spec.rb
```diff
@@ -0,0 +1,114 @@
+#
+# Author:: Bryan McLellan <btm@opscode.com>
+# Copyright:: Copyright (c) 2013 Opscode, Inc.
```
btm: Update the author/copyright in the header here.

spec/unit/knife/wsman_test_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require 'spec_helper'
+
+describe Chef::Knife::WsmanTest do
+  before(:all) do
```
btm: Chef::Config.reset was written for this purpose. We should put something in spec_helper.rb like we did in Chef

https://github.com/chef/chef/blob/master/spec/spec_helper.rb#L170

btm:
I see a bunch of trailing whitespace, could you check in with your editor about that? :)

lib/chef/knife/winrm_base.rb
smurawski: @btm Actually it doesn't.  If you drop into pry on a number of the unit tests off master (as it stands now), you'll see a bunch of config settings didn't resolve properly.

lib/chef/knife/winrm_knife_base.rb
smurawski: Done.  https://github.com/chef/chef/issues/3042

lib/chef/knife/winrm_base.rb
smurawski: I filed https://github.com/chef/knife-windows/issues/204 to track that.

btm:
I'm wondering a little bit about "knife wsman test" and "knife windows cert generate" and "knife windows listener create" -- once we move these into core, I wonder what we're going to want these commands to be; and can we prepare for that a little now?

smurawski:
@btm  wsman test is an accurate representation of what is happening.  It's validating the WSMan endpoint for the designated hosts.  While this is primarily for Windows nodes, WSMan endpoints are available for other platforms (via OMI or OpenPegasus for example) and on things like Cisco, Arista, and Dell switches (which have or will eventually DSC endpoints).

smurawski:
@btm Added.  Thanks - not sure how I missed those.

README.md
```diff
@@ -44,6 +44,15 @@ Or force a chef run:
 
 This subcommand operates in a manner similar to [knife ssh](https://docs.chef.io/knife_ssh.html)...just leveraging the WinRM protocol for communication. It also include's `knife ssh`'s "[interactive session mode](https://docs.chef.io/knife_ssh.html#options)"
 
+### knife wsman test
+
+Connects to the remote WSMan/WinRM endpoint and verifies the remote node is listening.  This is the equivalent of running Test-Wsman from PowerShell.  Endpoints to test can be specified manually, or be driven by search and use many of the same connection options as knife winrm.
+
+    knife wsman test 192.168.1.10 -m
```
btm: So --help prints a lot, the bulk of which isn't relevant most of the time, but maybe is periodically relevant.

An example like this would imply some good common flags:
```
 knife wsman test --manual-list 192.168.1.10 --winrm-port 443 --winrm-transport ssl
```

---
```knife wsman test --help```  output:
```
knife wsman test QUERY (options)
    -a, --attribute ATTR             The attribute to use for opening the connection - default is fqdn
    -f CA_TRUST_FILE,                The Certificate Authority (CA) trust file used for SSL transport
        --ca-trust-file
    -s, --server-url URL             Chef Server URL
        --chef-zero-host HOST        Host to start chef-zero on
        --chef-zero-port PORT        Port to start chef-zero on
    -k, --key KEY                    API Client Key
        --[no-]color                 Use colored output, defaults to false on Windows, true otherwise
    -c, --config CONFIG              The configuration file to use
        --defaults                   Accept default values for all questions
    -d, --disable-editing            Do not open EDITOR, just accept the data as is
    -e, --editor EDITOR              Set the editor to use for interactive commands
    -E, --environment ENVIRONMENT    Set the Chef environment (except for in searches, where this will be flagrantly ignored)
    -F, --format FORMAT              Which format to use for output
        --identity-file IDENTITY_FILE
                                     The SSH identity file used for authentication
    -i, --keytab-file KEYTAB_FILE    The Kerberos keytab file used for authentication
    -R KERBEROS_REALM,               The Kerberos realm used for authentication
        --kerberos-realm
    -S KERBEROS_SERVICE,             The Kerberos service used for authentication
        --kerberos-service
    -z, --local-mode                 Point knife commands at local repository instead of server
    -m, --manual-list                QUERY is a space separated list of servers
    -u, --user USER                  API Client Username
        --print-after                Show the data after a destructive operation
        --session-timeout Minutes    The timeout for the client for the maximum length of the WinRM session
    -V, --verbose                    More verbose output. Use twice for max verbosity
    -v, --version                    Show chef version
        --winrm-authentication-protocol AUTHENTICATION_PROTOCOL
                                     The authentication protocol used during WinRM communication. The supported protocols are basic,negotiate,kerberos. Default is 'negotiate'.
    -P, --winrm-password PASSWORD    The WinRM password
    -p, --winrm-port PORT            The WinRM port, by default this is '5985' for 'plaintext' and '5986' for 'ssl' winrm transport
        --winrm-ssl-verify-mode SSL_VERIFY_MODE
                                     The WinRM peer verification mode. Valid choices are [verify_peer, verify_none]
    -t, --winrm-transport TRANSPORT  The WinRM transport type.  valid choices are [ssl, plaintext]
    -x, --winrm-user USERNAME        The WinRM username
    -y, --yes                        Say yes to all prompts for confirmation
    -h, --help                       Show this message
```

File: knife-windows.gemspec [hunk 14]
```diff
@@ -15,7 +15,8 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version	= ">= 1.9.1"
   s.add_dependency "winrm", "~> 1.3"
-  s.add_dependency "winrm-s", "~> 0.2"
+  s.add_dependency "winrm-s", "~> 0.3.0.dev.0"
```
btm: maybe we should edit ```lib\knife-windows\version.rb``` to something like this:

```
    VERSION = "1.0.0.dev.0" # FIXME: release winrm-s 0.3.0 and update gemspec before release
```

We probably don't need to, and it's a hack. Maybe just typing this will make me remember :)

I'll make a card on our board.

btm: https://chef.leankit.com/Boards/View/157508079/177003065

btm:
This probably isn't necessary for a MVP, but I'd file an issue right away if we don't ship a fix for this. There's two bugs here:

1) It isn't super clear that we're successful until we try something that fails and we see the first run didn't have an error printed. That is, we should print something like "Successfully connected to host blah on port blah using protocol blah" and consider hiding the rest of the output unless -V

2) Success or failure doesn't change the exit result.

Success:
```
PS C:\Users\btm_000\Documents\github\knife-windows> knife wsman test --manual-list 192.168.5.109 --winrm-port 80
host:                 192.168.5.109
product_vendor:       Microsoft Corporation
product_version:      OS: 0.0.0 SP: 0.0 Stack: 1.1
protocol_version:     http://schemas.dmtf.org/wbem/wsman/1/wsman.xsd
response_status_code: 200
wsman_port:           80
wsman_url:            http://192.168.5.109:80/wsman
PS C:\Users\btm_000\Documents\github\knife-windows> echo $?
True
PS C:\Users\btm_000\Documents\github\knife-windows> echo $LASTEXITCODE
0
```

Failure:
```
PS C:\Users\btm_000\Documents\github\knife-windows> knife wsman test --manual-list 192.168.5.109 --winrm-port 443
error_message: No connection could be made because the target machine actively refused it. - connect(2) (http://192.168.5.109:443)
host:          192.168.5.109
wsman_port:    443
wsman_url:     http://192.168.5.109:443/wsman
PS C:\Users\btm_000\Documents\github\knife-windows> echo $?
True
PS C:\Users\btm_000\Documents\github\knife-windows> echo $LASTEXITCODE
0
```

smurawski:
@btm I get the idea of setting the error code in the case of a failure, but what should the desired behavior be in the case of testing multiple nodes?  Should one failure (out of x number of potential tests) flip the exit code in the same manner?

btm:
It looks like for knife ssh we return the largest exit code in the case of multiple nodes (https://github.com/chef/chef/commit/2fbdf2b2).

Yeah, I'd say we return an error if anything failed for multiple nodes.

smurawski:
Cool.  Thanks.  I'll look at making those changes.

smurawski:
@btm Fixed the output as you recommended and modified to exit with non-zero in the case of failures.

btm:
:+1: thanks bunches of bunches.

PR #150, opened 2014-12-22
knife bootstrap windows winrm fails with Bad HTTP error

NimishaS:
chef-client gets installed on the node but node doesn't get registered on the hosted chef account.


NimishaS:
Created separate card for @kaustubh-d 's suggestion.

PR #158, opened 2015-01-16
Fixed bug related to chef gem version check

siddheshwar-more:
```
DEBUG: Found bootstrap template in C:/opscode_live/chef/embedded/lib/ruby/gems/2.0.0/gems/knife-windows-1.0.0.dev.0/lib/chef/knife/bootstrap
ERROR: knife encountered an unexpected error
This may be a bug in the 'bootstrap windows winrm' knife command or plugin
Please collect the output of this command with the `-VV` option before filing a bug report.
Exception: NoMethodError: undefined method `version' for nil:NilClass
```

Implemented changes to fix above issue

### DIFF
--- a/lib/chef/knife/bootstrap/windows-chef-client-msi.erb
+++ b/lib/chef/knife/bootstrap/windows-chef-client-msi.erb
@@ -197,7 +197,7 @@ echo Validation key written.
 )
 <% end -%>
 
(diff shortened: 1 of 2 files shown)


siddheshwar-more:
@adamedx 
FYI - This will fix chef version check related issue.

PR #161, opened 2015-01-20
Updated Readme for winrm certgen commands

prabhu-das:


README.md
```diff
@@ -60,6 +60,34 @@ An initial run_list for the node can also be passed to the subcommand. Example u
 
     knife bootstrap windows ssh ec2-50-xx-xx-124.compute-1.amazonaws.com -r 'role[webserver],role[production]' -x Administrator -i ~/.ssh/id_rsa
 
+### knife windows cert generate
+
+Generate a certificate(x509) containing a public / private key pair for WinRM 'SSL' communication.
```
NimishaS: It should be "Generates a certificate(x509) containing public/private key pair for WinRM 'SSL' communication."

README.md
```diff
@@ -60,6 +60,34 @@ An initial run_list for the node can also be passed to the subcommand. Example u
...
+The certificate will be generated in three different formats *.pfx, *.der and *.pem.
+The PKCS12(i.e *.pfx) will be added to WinRM Server's Certificate Store.
+The *.der is base64 representaion of *.pfx which may be useful in cloud provider api.
+The *.pem contains public key and can be used at client side.
+This command also display's thumbprint of generated certificate.
```
NimishaS: "display's" should be "displays".

README.md
```diff
@@ -60,6 +60,34 @@ An initial run_list for the node can also be passed to the subcommand. Example u
...
+
+    knife windows cert generate --cert-passphrase "strong_passphrase" --domain "cloudapp.net" --output-file "C:\Users\Administrator\server_cert.pfx"
+    # This command will generate certificates at "C:\Users\Administrator\" location.
```
NimishaS: If this command generates certificate at the specified location with names as server_cert.pfx, server_cert.der and server_cert.pem, then we should mention about this in the command description

README.md
```diff
@@ -60,6 +60,34 @@ An initial run_list for the node can also be passed to the subcommand. Example u
...
+
+Generates a certificate(x509) containing a public / private key pair for WinRM 'SSL' communication.
+The certificate will be generated in three different formats *.pfx, *.der and *.pem.
+The PKCS12(i.e *.pfx) will be added to WinRM Server's Certificate Store.
+The *.der is base64 representaion of *.pfx which may be useful in cloud provider api.
+The *.pem contains public key and can be used at client side.
+This command also displays thumbprint of generated certificate.
+
+    knife windows cert generate --cert-passphrase "strong_passphrase" --domain "cloudapp.net" --output-file "C:\Users\Administrator\server_cert.pfx"
+    # This command will generate certificates at "C:\Users\Administrator\" location with names server_cert.der, server_cert.pfx and server_cert.pem.
```
adamedx: This command works on both Unix and Windows -- we should use an example that works in both cases. Assuming you're using PowerShell, you can use paths such as `~/server_cert.pfx` rather than `c:\users\administrator\server_cert.pfx`.

README.md
```diff
@@ -60,6 +60,34 @@ An initial run_list for the node can also be passed to the subcommand. Example u
...
+
+### knife windows cert install
+
+This command should be run on Windows Platform. The command will add the specified certificate to its certificate store. The command should include valid PKCS12(i.e *.pfx) certificate file path.
```
adamedx: I think a phrase like `This command only functions on Windows` is a little cleaner (same for the next command). Also, we should say **must** instead of *should* in the rest of the comments.

README.md
```diff
@@ -60,6 +60,34 @@ An initial run_list for the node can also be passed to the subcommand. Example u
...
+
+    knife windows cert install "C:\Users\Administrator\server_cert.pfx" --cert-passphrase "strong_passphrase"
+
+### knife windows listener create
+This command should be run on the Windows Platform. The command will create the winrm listener for SSL communication(i.e HTTPS).
+The command can also install certificate which is specified using --cert-install option and use installed certificate thumbprint to create winrm listener.
```
adamedx: `use installed` should be `use the installed`.

README.md
```diff
@@ -60,6 +60,34 @@ An initial run_list for the node can also be passed to the subcommand. Example u
...
+This command also displays thumbprint of generated certificate.
+
+    knife windows cert generate --cert-passphrase "strong_passphrase" --domain "cloudapp.net" --output-file "C:\Users\Administrator\server_cert.pfx"
+    # This command will generate certificates at "C:\Users\Administrator\" location with names server_cert.der, server_cert.pfx and server_cert.pem.
+
+### knife windows cert install
+
+This command should be run on Windows Platform. The command will add the specified certificate to its certificate store. The command should include valid PKCS12(i.e *.pfx) certificate file path.
+
+    knife windows cert install "C:\Users\Administrator\server_cert.pfx" --cert-passphrase "strong_passphrase"
```
adamedx: Would still be good to use `"~/server_cert.pfx"` since that's a valid syntax for PowerShell users.

README.md
```diff
@@ -60,6 +60,34 @@ An initial run_list for the node can also be passed to the subcommand. Example u
...
+### knife windows cert generate
+
+Generates a certificate(x509) containing a public / private key pair for WinRM 'SSL' communication.
+The certificate will be generated in three different formats *.pfx, *.der and *.pem.
+The PKCS12(i.e *.pfx) will be added to WinRM Server's Certificate Store.
+The *.der is base64 representaion of *.pfx which may be useful in cloud provider api.
+The *.pem contains public key and can be used at client side.
+This command also displays the thumbprint of the generated certificate.
+
+    knife windows cert generate --cert-passphrase "strong_passphrase" --domain "cloudapp.net" --output-file "~\server_cert.pfx"
```
adamedx: Just change the "\" to "/" since that works on powershell.

README.md
```diff
@@ -60,6 +60,34 @@ An initial run_list for the node can also be passed to the subcommand. Example u
 
     knife bootstrap windows ssh ec2-50-xx-xx-124.compute-1.amazonaws.com -r 'role[webserver],role[production]' -x Administrator -i ~/.ssh/id_rsa
 
+### knife windows cert generate
+
+Generates a certificate(x509) containing a public / private key pair for WinRM 'SSL' communication.
+The certificate will be generated in three different formats *.pfx, *.der and *.pem.
+The PKCS12(i.e *.pfx) will be added to WinRM Server's Certificate Store.
+The *.der is base64 representaion of *.pfx which may be useful in cloud provider api.
+The *.pem contains public key and can be used at client side.
```
btm: somethings wrong here with the generated files, and these descriptions. I'll open another issue about the generated files.

The der file should be a binary format copy of the certificate, which is primary a Windows format (maybe Azure likes it?)
The pem file should be a base64 (text readable) copy of the certificate
The pfx file should contain both the certificate and the private key it is signed with.

README.md
```diff
@@ -60,6 +60,34 @@ An initial run_list for the node can also be passed to the subcommand. Example u
...
+
+    knife windows cert install "~\server_cert.pfx" --cert-passphrase "strong_passphrase"
+
+### knife windows listener create
+This command only functions on Windows. It creates the winrm listener for SSL communication(i.e HTTPS).
+This command can also install certificate which is specified using --cert-install option and use the installed certificate thumbprint to create winrm listener.
+
+    knife windows listener create --cert-passphrase "strong_passphrase" --hostname "*.cloudapp.net" --cert-install "~\server_cert.pfx"
+
+The command also allows you to use existing certificates from local store to create winrm listener. Use --cert-thumbprint option to specify the certificate thumbprint.
```
btm: It'd be helpful to include how to get the thumbprint here if we didn't save it from when we generated it.

README.md
```diff
@@ -60,6 +60,34 @@ An initial run_list for the node can also be passed to the subcommand. Example u
...
+
+    knife windows listener create --cert-passphrase "strong_passphrase" --hostname "*.cloudapp.net" --cert-thumbprint "bf0fef0bb41be40ceb66a3b38813ca489fe99746"
```
btm: Hostname appears optional, what should we be putting there?

README.md
btm: Filed the der issue as #163

btm: Fixes in #165

Issue #163, opened 2015-01-23
knife windows cert generate - der file appears broken


btm:
The der file should be a binary file, it looks like a base64 encoded key without the headers that you would see with a pem file.

NimishaS:
@btm , can you please review https://github.com/chef/knife-windows/pull/165 ?

btm:
Fixed by #165.

PR #165, opened 2015-01-28
Changed .der extension to .b64

NimishaS:
Fix for issue https://github.com/chef/knife-windows/issues/163

Azure requires base64 encoded `pfx` file: https://msdn.microsoft.com/en-us/library/azure/ee460817.aspx
Since we were using `der` file for that purpose only, we should make the extension accurate.

### DIFF
--- a/lib/chef/knife/windows_cert_generate.rb
+++ b/lib/chef/knife/windows_cert_generate.rb
@@ -118,9 +118,9 @@ class Chef
           cert = generate_certificate rsa_key
           write_certificate_to_file cert, file_path, rsa_key
           ui.info "Generated Certificates:"
-          ui.info " PKCS12 FORMAT (needed on the server machine, contains private key): #{file_path}.pfx"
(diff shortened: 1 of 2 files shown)


NimishaS:
We referred to various blogs regarding certificate:
https://myonlineusb.wordpress.com/2011/06/19/how-to-convert-certificates-between-pem-der-p7bpkcs7-pfxpkcs12/
http://blogs.msdn.com/b/kaushal/archive/2010/11/05/ssl-certificates.aspx
https://support.ssl.com/Knowledgebase/Article/View/19/0/der-vs-crt-vs-cer-vs-pem-certificates-and-how-to-convert-them

Description of pem file(in 1st link) says that it is a base64 encoded file and can contain certificate and private key as well. It can have extension as .pem, .crt, .cer, .key but should contain the `—–BEGIN CERTIFICATE—–` and `—–END CERTIFICATE—–` statements. We thought of using one of these extensions but since our file doesn't have `—–BEGIN CERTIFICATE—–` and `—–END CERTIFICATE—–`, we are using `b64` extension

PR #166, opened 2015-01-29
Updated Readme for setting MaxTimeoutms for windows 2008r2

NimishaS:
Windows 2008r2 and earlier require an extra configuration.

We should set  MaxTimeoutms >= 30000 to avoid WinRM::WinRMHTTPTransportError: Bad HTTP response error while bootstrapping

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,7 +169,7 @@ For development and testing purposes, unencrypted traffic with Basic authenticat
 
     winrm set winrm/config/winrs '@{MaxMemoryPerShellMB="512"}'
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -136,6 +136,10 @@ For development and testing purposes, unencrypted traffic with Basic authenticat
 
     winrm set winrm/config/winrs '@{MaxMemoryPerShellMB="512"}'
 
+* Windows 2008r2 and earlier versions require an extra configuration for MaxTimeoutms to avoid WinRM::WinRMHTTPTransportError: Bad HTTP response error while bootstrapping. It should be greater than 30000.
```
adamedx: Probably better to say at least 30000 since the example sets it to 30000.

NimishaS: Note: 30000 is incorrect here. It should be 300000.
300000 ms = 300 s

PR #167, opened 2015-01-29
Update winrm_authentication_protocol option

prabhu-das:
Changes to display default values for winrm_authentication_protocol option in CLI help.


adamedx:
Ok, so I think maybe the doc is right, but we should change how this works. :smile: .

Here's how I think it should work
1. Negotiate should always be the default
2. If you use plaintext, we should use encryption via SSPI (the winrm-s patch) -- if you're not on Windows we should warn since there is probably no support in GSSAPI to encrypt

I don't think you can use basic by default since by default winrm listeners don't support basic auth

btm:
:+1: to @adamedx's comment.

PR #168, opened 2015-02-04
Remove domain name and Negotiate protocol validations

siddheshwar-more:
Removed domain related validation with respective negotiate protocol and fixed/added unit tests.


adamedx:
Looks good -- can you add a spec that validates that we don't try the monkeypatch if you use the ssl transport?

siddheshwar-more:
@adamedx 
Fixed review comment.
Can we merge this?
Thanks!

adamedx:
:+1: -- Thanks Siddheshwar.

PR #185, opened 2015-02-11
Update Rubies, removing unsupported 1.9.x

juliandunn:
This will fix the tests running in Travis


btm:
cherry-picked to 20356229d1cf9b34ae304af50685bf26f598e52c as well, which hopefully goes away after 1.0.0 or so (soon).

PR #186, opened 2015-02-11
Add connection_timeout property that defaults to 30 minutes but can be overrid...

carpnick:
...den on the CLI.

This PR along with the new winrm-s [PR \#22](https://github.com/chef/winrm-s/pull/22) and issue[https://github.com/chef/winrm-s/issues/21] will get knife-windows to have a configurable timeout for both the operation timeout and receive_timeout.

Both of these are required to enable greater than 60 minute winrm sessions  For reference the WinRM 1.3 PR(https://github.com/WinRb/WinRM/pull/108)

PS -
Depending on the integration of the new WinRM in winrm-s gem, we might need to do one more commit here in the gemspec file...

### DIFF
--- a/lib/chef/knife/winrm.rb
+++ b/lib/chef/knife/winrm.rb
@@ -185,7 +185,7 @@ class Chef
           session_opts[:service] = locate_config_value(:kerberos_service) if locate_config_value(:kerberos_service)
           session_opts[:ca_trust_path] = locate_config_value(:ca_trust_file) if locate_config_value(:ca_trust_file)
           #30 min (Default) OperationTimeout for long bootstraps fix for KNIFE_WINDOWS-8
-          session_opts[:operation_timeout] = locate_config_value(:connection_timeout).to_i if locate_config_value(:connection_timeout)
+          session_opts[:operation_timeout] = locate_config_value(:session_timeout).to_i * 60 if locate_config_value(:session_timeout)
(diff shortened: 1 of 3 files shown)


btm:
I wonder if this should just be "timeout". I was surprised we didn't have something similar for knife bootstrap. There is a "knife ec2 --windows-auth-timeout". A "connection timeout" makes me think it is just for establishing the connection, whereas the length of the connection I'd call a session timeout. 

Since the default is 60 minutes, I don't see much value in specifying seconds instead of minutes. If you need more time you're going to want 120 minutes, not 60 minutes and 1 second / 1801 seconds. The aforementioned timeout in EC2 is in minutes too.

As a final though, since it's so long, maybe it should be a --session-timeout option, in minutes.

carpnick:
Thanks for the review @btm.  I will switch it to  --session-timeout option, in minutes..

The default currently is actually 30 minutes.  But the WinRM gem had a max time before 1.3 of 60 minutes for the http (winrm) session.

carpnick:
Build failed due to unrelated issue in appveyor with httpclient.  Not related to this change...

carpnick:
@btm - let me know if you need anything else.

jdmundrawala:
:+1:

ksubrama:
Rebased and merged in 0d8b621cca376399825d0df1c321052ed4da220c

PR #187, opened 2015-02-11
Added new option for cert generate command to overwrite existing certs

prabhu-das:
Added changes to support- cert generate command overwrites existing certs if destination files already exist.
Introduced new option `--overwrite-certs`

### DIFF
--- a/lib/chef/knife/windows_cert_generate.rb
+++ b/lib/chef/knife/windows_cert_generate.rb
@@ -61,6 +61,12 @@ class Chef
         :long => "--cert-passphrase PASSWORD",
         :description => "Password for certificate."
 
+      option :overwrite_certs,
+        :long => "--overwrite-certs",
+        :boolean => true,
+        :description => "The existing certificates files will be overwritten, if the --output-file <file_path> that the cert generate command creates already exist.",
+        :default => false
+
       def generate_keypair
         OpenSSL::PKey::RSA.new(config[:key_length].to_i)
       end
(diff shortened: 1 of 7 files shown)


lib/chef/knife/windows_cert_generate.rb
```diff
@@ -56,6 +56,12 @@ class WindowsCertGenerate < Knife
         :long => "--cert-passphrase PASSWORD",
         :description => "Password for certificate."
 
+      option :overwrite_certs,
```
adamedx: I actually don't think we need this flag -- it's something we could add back later if users ask for it.

btm:
Agree with @adamedx. If anything we'd probably roll this functionality into something like a "--force" flag.

Although, we should probably raise an error and exit if the files exist. It looks like we're blindly overwriting them right now.

adamedx:
Yeah, I think there's a line that does do what you suggest @btm -- overwrite_existing_certs! But we could probably just change the name of that method and also have it not read from config since we are getting rid of the config option.

prabhu-das:
@btm @adamedx 
Fixed review comments.
Thank you.

lib/chef/knife/windows_cert_generate.rb
```diff
@@ -108,11 +108,32 @@ def write_certificate_to_file cert, file_path, rsa_key
         File.open(file_path + ".b64", "wb") { |f| f.print Base64.strict_encode64(pfx.to_der) }
       end
 
+      def is_other_certificates_present?(file_path)
+        is_certs_exists = false
```
NimishaS: Suggestion for method and variable names:
1. `is_other_certificates_present?` method name can be `certificates_already_exist?`
2. `is_certs_exists` variable name can be `certs_exist`

adamedx: Agree with @NimishaS's feedback.

prabhu-das: Fixed comments.
Thanks!

PR #190, opened 2015-02-12
Fix badge status links

jdmundrawala:


jdmundrawala:
Merging as a trivial change

Issue #148, opened 2014-12-19
escape_glob missing for Chef::Util::PathHelper::Class


usepowershell:
When using 1.0.0.dev.0 with ChefDK 3.5 on a Mac - I get a failure to bootstrap a windows node due to a missing 'escape_glob' in Chef::Util::PathHelper::Class.

The full -VV output is here - 

```
smurawski@remmurawski01:chef-repo$ knife bootstrap windows winrm armory -x Administrator -P WalkerBlack!

Waiting for remote response before bootstrap.armory Response received.
armory 
Remote node responded after 0.01 minutes.
Bootstrapping Chef on armory
ERROR: knife encountered an unexpected error
This may be a bug in the 'bootstrap windows winrm' knife command or plugin
Please collect the output of this command with the `-VV` option before filing a bug report.
Exception: NoMethodError: undefined method `escape_glob' for Chef::Util::PathHelper:Class
smurawski@remmurawski01:chef-repo$ knife bootstrap windows winrm armory -x Administrator -P Passw@rd! -VV

Waiting for remote response before bootstrap.DEBUG: Adding armory
DEBUG: WinRM::WinRMWebService options: {:user=>"Administrator", :pass=>"Passw@rd!", :basic_auth_only=>true, :disable_sspi=>true, :no_ssl_peer_verification=>false, :ca_trust_path=>nil}
DEBUG: Endpoint: http://armory:5985/wsman
DEBUG: Transport: plaintext
DEBUG: echo . & echo Response received.
armory Response received.
armory 
Remote node responded after 0.0 minutes.
Bootstrapping Chef on armory
DEBUG: Looking for bootstrap template in /Users/smurawski/.chefdk/gem/ruby/2.1.0/gems/knife-windows-1.0.0.dev.0/lib/chef/knife/bootstrap
DEBUG: Found bootstrap template in /Users/smurawski/.chefdk/gem/ruby/2.1.0/gems/knife-windows-1.0.0.dev.0/lib/chef/knife/bootstrap
ERROR: knife encountered an unexpected error
This may be a bug in the 'bootstrap windows winrm' knife command or plugin
Please collect the output of this command with the `-VV` option before filing a bug report.
Exception: NoMethodError: undefined method `escape_glob' for Chef::Util::PathHelper:Class
```

btm:
Chef DK 0.4.0, released on 2/9/15 has Chef 12 as the tools version now, so this shouldn't hit us there but we should test Chef 11 still because many people use it.

[...]

PR #191, opened 2015-02-13
Use ::Chef::Util::PathHelper since the new Knife::Windows::PathHelper isn't released yet

jeremiahsnapp:


btm:
Chef::Util::PathHelper.escape_glob never made it into a Chef 11 release, which is why we were going to vendor PathHelper in knife-windows=1.0.0 (so it would work with ChefDK 0.3.5, which had Chef 11 as the tool version). Now that ChefDK 0.4.0 is out which fully uses Chef 12, it's less of an issue, but I think we still want to maintain Chef 11 support for now. In any case, https://github.com/chef/knife-windows/issues/148 has already been merged into 0.8-release and shipped as 0.8.4.rc.4 to ensure Chef 11 compatibility in 0.8.4. I think we need to maintain Chef 11 compatibility for now in 0.8.4 since we want to ship it immediately.

jeremiahsnapp:
@btm :+1: I wasn't aware of that. I'm glad you caught it.

0.8.4.rc.4 hasn't been released yet so I assume that's a typo. It's not clear to me what the next steps are to get a working 0.8.4.rc released. Is there anything I need to do?

btm:
@jeremiahsnapp yeah typo I edited it to 0.8.4.rc.3 a bit ago. trying to type fast to keep everyone up to date is error-prone apparently.  0.8.4.rc.3 is up.

PR #195, opened 2015-02-26
Pass `ssl_verify_mode` bootstrap config to WinRM command

schisamo:
/cc @adamedx @btm


adamedx:
Thank you @schisamo. Separately, @NimishaS, we should look at finding a way to migrate "universal" options into the plugins. Perhaps that's just by using knife-cloud. Or we could have an option for the WinRM bootstrapper to just initialize itself from config.

adamedx:
:+1:

PR #196, opened 2015-02-26
Fix trusted_certs regression introduced by c09a2b2

schisamo:
The helper for accessing trusted_certs content was renamed to `trusted_certs_script` in 583a2b2. For some reason this was changed back to `trusted_certs` in commit c09a2b2. Without this fix bootstrapping fails with:

    NameError: undefined local variable or method `trusted_certs' for #<Chef::Knife::Core::WindowsBootstrapContext:0x007fcd54522910>

/cc @adamx @btm


adamedx:
:+1:

PR #197, opened 2015-02-27
WinRM 1.3.0 compat

schisamo:
Test fixes to make knife-windows compatible with the recently released [WinRM 1.3.0](Fixed multiple issues with WinRMHTTPTransportError incorrectly being raised). In my testing the new version has proven much more stable especially WRT `WinRMHTTPTransportError` exceptions being incorrectly raised.

/cc @adamedx @btm @smurawski


adamedx:
Thanks for the fix @schisamo.

PR #198, opened 2015-02-27
knife-windows WinRM Listener Create should default to blank hostname if not specified or allow --nohostname to be specified.

prabhu-das:
Updated hostname default option to blank for knife windows listner create and removed * as default hostname


lib/chef/knife/windows_listener_create.rb
```diff
@@ -40,7 +40,7 @@ class WindowsListenerCreate < Knife
         :short => "-h HOSTNAME",
         :long => "--hostname HOSTNAME",
         :description => "Hostname on the listener. Default is *",
```
NimishaS: Description should be changed as well.

NimishaS:
Default value of hostname(common name) is "*" while certificate generation. Should it be consistent with this behavior?

adamedx:
I think * is not appropriate in either case.

btm:
The hostname used in cert generate goes into the CN, right? I would expect it has to be something and we should raise an error if it isn't provided. I'm presuming winrm is going to be matching the hostname against the CN and give us an error if it doesn't match.

Have we tested connecting to a system with a certificate with a CN of * or blank?

NimishaS:
@btm , yes hostname in cert generate goes to CN. @adamedx had suggested to set it to local fqdn when not provided.

We tested that connecting to a system works doesn't work when CN is blank.

adamedx:
Correct, we'll assume a "local" workflow where you are trying to set up the listener for the system you're running on. In that case, you want the cert to be the fqdn (assuming fqdn is also the server name in the listener url, which should be common).

Otherwise, you need to set this yourself to match the url of the listener.

btm:
Taking https://github.com/chef/knife-windows/pull/201 into consideration, we should mention CN because that's meaningful to a lot of people who know something or are reading something about certificates.

```
:description => "Use to specify the remote hostname for the CN field of the certificate.
    For example, --hostname windows_node.example.org or *.example.org
    Default value is the hostname of the local system."
```

NimishaS:
@btm, we'll fix the description in a separate PR where we set default hostname of certificate to local fqdn.

NimishaS:
Something got wrong during rebase. Created https://github.com/chef/knife-windows/pull/205 cherry-picking the changes from this PR.

Issue #193, opened 2015-02-23
encrypted_data_bag_secret is the same as secret_file option in knife


adamedx:
We should not have different names for the same option. We should change the gem to use either option, and if both are specified, it should fail with a warning saying to use the more generic option from knife. We should do this for all conflicting options.

btm:
Fixed in master by #199

PR #199, opened 2015-03-02
Added chef options consistency changes

prabhu-das:
options changes- encrypted_data_bag_secret and encrypted_data_bag_secret_file to secret & secret_file respectively.

Issue Reference- https://github.com/chef/knife-windows/issues/193

### DIFF
--- a/lib/chef/knife/bootstrap/windows-chef-client-msi.erb
+++ b/lib/chef/knife/bootstrap/windows-chef-client-msi.erb
@@ -191,9 +191,9 @@ echo Writing validation key...
 echo Validation key written.
 @echo on
 
-<% if @config[:encrypted_data_bag_secret] -%>
+<% if @config[:secret] -%>
 > <%= bootstrap_directory %>\encrypted_data_bag_secret (
- <%= encrypted_data_bag_secret %>
+ <%= secret %>
 )
 <% end -%>
 
--- a/lib/chef/knife/bootstrap_windows_base.rb
+++ b/lib/chef/knife/bootstrap_windows_base.rb
@@ -85,13 +85,13 @@ class Chef
             :proc => lambda { |o| JSON.parse(o) },
             :default => {}
 
-          option :secret,
+          option :encrypted_data_bag_secret,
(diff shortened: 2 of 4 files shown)


btm:
There was a warning in core Chef about the deprecation, add in 11.10.0 (https://github.com/chef/chef/commit/6c684ef5) and removed in 12.0.0 (https://github.com/chef/chef/commit/63c04487).

So it seems like there's been plenty of time for people to switch, but Windows users had to keep using the old values here. The remove of the command line option will be a pretty obvious breaking change; that is, it won't work, hopefully people will run a --help, and see that it changed. For the knife config, hopefully they've seen this coming because of the former warning.

But I still feel like we should keep the old options but throw an error, and probably throw an warning if we detect ```config[:encrypted_data_bag_secret]``` or ```config[:encrypted_data_bag_secret_file]``` are set.

lib/chef/knife/bootstrap/windows-chef-client-msi.erb
```diff
@@ -191,9 +191,9 @@ echo Writing validation key...
 echo Validation key written.
 @echo on
 
-<% if @config[:encrypted_data_bag_secret] -%>
+<% if @config[:secret] -%>
 > <%= bootstrap_directory %>\encrypted_data_bag_secret (
- <%= encrypted_data_bag_secret %>
```
adamedx: Won't this break people who are using encrypted_data_bag_secret already? We should use `secret` instead of encrypted_data_bag_secret, and warn if `encrypted_data_bag_secret` is specified that they should use `secret` instead, so we can deprecate it later. We might even be able to use deprecation support built into Chef.

prabhu-das: @adamedx  We have taken care of above in this PR Please review : https://github.com/ClogenyTechnologies/knife-windows/commit/b57512bca48015b6d92a4dc320056737d36ce9df

NimishaS:
@btm , please review this.

NimishaS:
@btm , should I merge this?

Issue #200, opened 2015-03-02
windows_cert_generate.rb


jamescott:
The description for --hostname does not have gender-neutral language in the description.

jamescott:
Specifically: "If user wants to give his fqdn as the hostname then he must specify the hostname as: --hostname 'something.mydomain.com'."

Should be edited to remove the "he" and "his".

btm:
fixed by #201

PR #201, opened 2015-03-03
Gender neutral message

NimishaS:
wrt: https://github.com/chef/knife-windows/issues/200

### DIFF
--- a/lib/chef/knife/windows_cert_generate.rb
+++ b/lib/chef/knife/windows_cert_generate.rb
@@ -31,9 +31,8 @@ class Chef
       option :hostname,
         :short => "-H HOSTNAME",
         :long => "--hostname HOSTNAME",
-        :description => "By default hostname is *. If user wants to give an fqdn as the hostname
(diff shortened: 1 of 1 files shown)


matt-richardson:
Can I suggest second person tense instead? That will match with other messages in the file:
````
By default hostname is *. If you want to give an fqdn as the hostname, then specify the hostname
as: --hostname 'something.mydomain.com'. You can also give wildcard as '*.mydomain.com'
````

jamescott:
This is how I wrote this for chef-docs:

"Use to specify the hostname for the listener. For example, --hostname something.mydomain.com or *.mydomain.com. Default value: *.

That said, we're not as formal within the descriptions in the code itself.

So the suggestion by @matt-richardson to use "you" is OK too.

Basically, we want to avoid using pronouns as a general rule, but definitely avoid any use of gender-specific pronouns. He, she, he/she, him, her, himself, herself should be avoided. You is more general and is perfectly OK if that's what the sentence needs.

NimishaS:
@jamescott , please review this.

Issue #192, opened 2015-02-13
deprecate knife bootstrap --distro


btm:
From lib/chef/knife/bootstrap.rb:
```
      # DEPR: Remove this option in Chef 13
      option :distro,
        :short => "-d DISTRO",
        :long => "--distro DISTRO",
        :description => "Bootstrap a distro using a template. [DEPRECATED] Use -t / --bootstrap-template option instead.",
        :proc        => Proc.new { |v|
          Chef::Log.warn("[DEPRECATED] -d / --distro option is deprecated. Use -t / --bootstrap-template option instead.")
          v
        }
```

The above deprecation is from https://github.com/chef/chef/pull/1885. We should deprecate both --distro --template in knife-windows in lib/chef/knife/bootstrap_windows_base.rb.

Which is easy enough, except theres some inconsistency in --bootstrap-template in knife-windows.

again from lib/chef/knife/bootstrap.rb:
```
      option :bootstrap_template,
        :short => "-t TEMPLATE",
        :long => "--bootstrap-template TEMPLATE",
        :description => "Bootstrap Chef using a built-in or custom template. Set to the full path of an erb template or use one of the built-in templates."
```

We need to add this option to knife-windows as the preferred option as well.

btm:
Looking at this again right now, I don't think we need --distro in knife-windows.

But we need to be consistent with the template option. Both ```knife bootstrap windows winrm``` and ```knife bootstrap windows ssh``` should have a --template-file option that matches ```knife bootstrap``` and we should deprecate --bootstrap-template and -t in favor of that (throw a warning to switch).

siddheshwar-more:
@btm 
In chef master https://github.com/chef/chef/blob/master/lib/chef/knife/bootstrap.rb#L130-L137 `--template-file` option is deprecated and  its using `--bootstrap-template` option.

but in above comment says-  we should deprecate `--bootstrap-template` option and `-t` 
which is little misleading here, Could you please clear our confusion about which option should we deprecate for knife-windows ? 

Thank  you!

btm:
If you look in chef master for knife bootstrap, both --distro and --template-file are deprecated, leaving --bootstrap-template/-t  as the preferred option.

In [knife-windows master](https://github.com/chef/knife-windows/blob/master/lib/chef/knife/bootstrap_windows_base.rb) we have --template-file and --distro.  To be consistent, I believe we need to add a --bootstrap-template/ -t, and to add depreciation messages to the others in knife windows.

[...]

PR #206, opened 2015-03-09
Add a flag msi_url that allows one to fetch the Chef client msi from a non-chef.io path.

ksubrama:


lib/chef/knife/bootstrap_windows_base.rb
```diff
@@ -114,6 +114,12 @@ def self.included(includer)
             :long        => "--[no-]node-verify-api-cert",
             :description => "Verify the SSL cert for HTTPS requests to the Chef server API.",
             :boolean     => true
+
+          option :msi_url,
+            :short => "-u URL",
+            :long => "--msi_url URL",
+            :description => "Location of the Chef Client MSI. The default templates will prefer to download from this location. The MSI will be downloaded from chef.io if not provided.",
```
btm: We don't fallback, to another location, so there isn't a download preference to prefer, right?
```
Location of the Chef Client MSI to be installed by the bootstrap template. Default is to download from chef.io.
```

lib/chef/knife/core/windows_bootstrap_context.rb
```diff
@@ -161,10 +161,27 @@ def latest_current_windows_chef_version_query
         end
 
         def win_wget
+          # I tried my best to figure out how to properly url decode and switch / to \
+          # but this is VBScript - so I don't really care that badly.
           win_wget = <<-WGET
 url = WScript.Arguments.Named("url")
 path = WScript.Arguments.Named("path")
 proxy = null
+If InStr(url, "file://") = 1 Then
```
btm: Take the above comment and put in in the VBS here with something like this, as up above the context is a little confusing, especially when vbscript just looks like random text to you. ;)
```
'* Handle file:// paths, but doesn't handle converting forward slashes to backslashes.
```

lib/chef/knife/bootstrap/windows-chef-client-msi.erb
```diff
@@ -107,12 +107,15 @@ goto install
...
-<% url="https://www.chef.io/chef/download?p=windows&pv=%MACHINE_OS%&m=%MACHINE_ARCH%" -%>
-<% url += latest_current_windows_chef_version_query -%>
-@set "REMOTE_SOURCE_MSI_URL=<%= url %>"
+<% default_url = "https://www.chef.io/chef/download?p=windows&pv=%MACHINE_OS%&m=%MACHINE_ARCH%" -%>
+<% default_url += latest_current_windows_chef_version_query -%>
+<% cs_url = msi_url.empty? ? default_url : msi_url -%>
+<% ps_url = msi_url.empty? ? default_url + "&DownloadContext=PowerShell" : msi_url -%>
+
+@set "REMOTE_SOURCE_MSI_URL=<%= cs_url %>"
+@set "REMOTE_SOURCE_MSI_PS_URL=<%= ps_url %>"
```
btm: there are a few places where we put %REMOTE_SOURCE_MSI_URL% into log output when something fails. I don't know why we pass "&DownloadContext=PowerShell" but maybe it would be important if that made it to the log output too.

For example, someone could say a URL is failing for them, and we try it and we can't reproduce it, but that's because REMOTE_SOURCE_MSI_PS_URL is being used but REMOTE_SOURCE_MSI_URL is being printed.

lib/chef/knife/core/windows_bootstrap_context.rb
```diff
@@ -238,6 +256,10 @@ def local_download_path
           local_download_path = "%TEMP%\\chef-client-latest.msi"
         end
 
+        def msi_url
```
btm: It seems like some of the logic could be moved in here from windows-chef-client-msi.erb, so if anyone else uses a custom template and needs our URL stuff they wouldn't be hardcoding it?

lib/chef/knife/core/windows_bootstrap_context.rb
```diff
@@ -238,6 +256,10 @@ def local_download_path
...
+          msi_url = @config[:msi_url] || ''
```
btm: how about just ```@config[:msi_url] || ''```? 

Or if the default for @config[:msi_url] was not set, even just ```@config[msi_url]```

or my other comment I guess.

lib/chef/knife/bootstrap/windows-chef-client-msi.erb
ksubrama: Hmmm...  well currently, if they depended on it, the log would show the wrong thing because &DownloadContext is not added to REMOTE_SOURCE_MSI_URL - it's just being dumped into the url during the ps invocation below.  I'll try to refactor this so that REMOTE_SOURCE_MSI_URL is being set to the right thing after we decide if we are using PS or not.

lib/chef/knife/bootstrap_windows_base.rb
ksubrama: I don't believe there is any fallback if the url is specified.  We only go to chef.io is nothing is specified.

ksubrama:
@btm I've fixed the issues pointed out above.  This is ready for another look.

btm:
:+1:

@carpnick I didn't catch if you use an alternate template for more than specifying the MSI URL. If that's all you need, you might want to keep an eye on this. It will be released with knife-windows 1.0.0.

btm:
@ksubrama I split out https://github.com/chef/knife-windows/issues/192 into ECOSYSTEM-293 under "tech debt" so we didn't lose it in the comments of this work.

carpnick:
Thanks @btm.  Actually use it also for installing a workaround to executing chef via winrm.  But thanks for the heads up

PR #208, opened 2015-03-11
Setting --hostname default value to local system's hostname for knife windows certificate generation

prabhu-das:


lib/chef/knife/windows_cert_generate.rb
```diff
@@ -75,7 +76,7 @@ def prompt_for_passphrase
       end
 
       def generate_certificate rsa_key
-        @hostname = "*"
+        @hostname = Socket.gethostname
```
btm: Have we tried this? If it doesn't work for basic testing I'm not convinced it's a worthwhile default. What's the use case here, is someone going to run 'knife windows cert generate' then 'knife windows listener create' then 'knife bootstrap windows winrm localhost'?

That doesn't make sense, I maybe we're expecting the person to run 'knife windows cert generate' on the server that they're going to connect to? In that case, this just returns hostname, not fqdn. In my experience with HTTPS, you're almost always connecting to the fqdn. Is winrm any different? If we really want FQDN, we can't get it from Socket (without looking up reverse dns, which isn't what we want).

Vasu1105: @btm 
Socket.gethostname will always pickup hostname. 
we made this changes for following use case-
1. User will run the windows cert generate command on remote server to create ssl cert
2. If --hostname is not specified then by default it will take server 'hostname' (i.e retrived  by using Socket.gethostname).
But now we realize generated cert is not working with hostname, it needs server's 'fqdn'.
And to find out 'fqdn' we can take help of reverse dns lookup. but is it right way to get fqdn? As i think dns reverse lookup is seems quite tricky!

or we should just make this fqdn option manadatory?

btm: I'm feeling like we should make it mandatory and not have a default. The user knows what the fqdn is that they're going to try to connect to right after they make this certificate, which is our best chance to be right most of the time with the least trouble.

I just looked at what [ohai does to get fqdn on windows](https://github.com/chef/ohai/blob/master/lib/ohai/plugins/hostname.rb#L163) and on a couple of my systems it just returns the hostname, even for fqdn. Doing a reverse DNS lookup requires that the user 1) has DNS correctly set up and 2) that the reverse DNS matches what it should. It's really common for reverse DNS to return nothing (or a generalized name based on ip address), in which case we would want to be sure to fail early. The hostname that the user is connecting to to get to the server isn't necessarily the reverse DNS either, e.g. reverse dns returns unique-asset-id.example.org, but I'm connecting to service-name.example.org.

adamedx: It's definitely your call @btm -- we can always add a default later if we decide that's helpful. My last feedback on this is that on Windows systems, DNS is almost always Active Directory, and particularly for domain-joined systems, the fqdn can be obtained very reliably. Also, Windows users often *don't* really know what the fqdn is, particularly in domains that are actually subdomains (they don't always know they're in a subdomain) and not the root of a forest.

Regardless, I think we can give good instructions here and if we find data indicating this is confusing (e.g. mailing list questions, etc.), we can revisit -- adding a default wouldn't break compatibility.

Thanks for looking at this!

adamedx: Minor nit: on Windows, you don't need reverse dns to get the fqdn, for domain-joined systems under the covers you just call [GetComputerNameEx](https://msdn.microsoft.com/en-us/library/windows/desktop/ms724301(v=vs.85).aspx) with name format `ComputerNameDnsFullyQualified`. Or call DsCrackNames with a cached name value obtained from WMI or other sources. In one case, it's an LDAP call to the DC (I think) that is cached so you actually don't have to go off the machine and the other also involves no network interaction.

btm: Ergh. I think I still have an AD controller here I could test this with, but hopefully you can answer this off the top of your head:

Say I have a server. web01.atlanta.eng.contoso.com. As a Windows administrator who might not know where I am in the forest, am I going to run ```knife windows bootstrap winrm web01.atlanta``` because I usually type web01.atlanta into remote desktop? Is that going to work if my CN is web01.atlanta.eng.contoso.com? Or would I need a bunch of SANs for that?

btm:
Replaced by #219

PR #212, opened 2015-03-23
Fix knife bootstrap check scripts to be a little more resilient.

ksubrama:


ksubrama:
/cc @btm @jdm @smurawski 

This was simply an exploratory attempt to fix some problems I'd seen in the bootstrap script before I realized it was more pain than it was worth it to really fix all the issues.  I documented the issues and have a partial fix.

jdm:
You want @jdmundrawala, not me.

lib/chef/knife/bootstrap_windows_base.rb
```diff
@@ -199,20 +199,46 @@ def bootstrap_command
...
+      end
+
+      def create_bootstrap_bat_command()
         chunk_num = 0
+        bootstrap_bat = ""
+        banner = bootstrap_render_banner_command(chunk_num += 1)
         render_template(load_template(config[:bootstrap_template])).each_line do |line|
           # escape WIN BATCH special chars
+          # TODO: The commands are going to get redirected - do we need to escape &?
           line.gsub!(/[(<|>)^]/).each{|m| "^#{m}"}
```
smurawski: This is a bit of a pet peeve of mine.. if you need a comment to explain a line (or two), why not extract a method named appropriately?  Then we are less likely to suffer comment rot and the intent will be clear in the source method.
Example:
```ruby
def escape_special_windows_batch_characters(line)
  # TODO: The commands are going to get redirected - do we need to escape &?
  line.gsub(/[(<|>)^]/).each{|m| "^#{m}"}
end
```

Some of the other comments lower in the method are great - explaining why, rather than what.  Comments for the why are more important as they help build context.

smurawski:
:+1:

lib/chef/knife/bootstrap_windows_base.rb
ksubrama: Good call.  Sometimes I get lazy and I appreciate someone poking me to do the right thing :)

PR #213, opened 2015-03-23
Ksubrama/env fix

ksubrama:


ksubrama:
Use the path-helper code from core chef client to discover bootstrap templates.

@btm @jdm @smurawski

jdm:
You want @jdmundrawala, not me.

btm:
@ksubrama can you put context onto this PR? I forget what this fixes.

lib/knife-windows/path_helper.rb
```diff
@@ -148,6 +148,83 @@ def self.escape_glob(*parts)
...
+          # HOMEDRIVE HOMEPATH
+          # HOMESHARE HOMEPATH
+          # USERPROFILE
+
+          paths << ENV['HOME']
+          paths << ENV['HOMEDRIVE'] + ENV['HOMEPATH'] if ENV['HOMEDRIVE'] && ENV['HOMEPATH']
+          paths << ENV['HOMESHARE'] + ENV['HOMEPATH'] if ENV['HOMESHARE'] && ENV['HOMEPATH']
+          paths << ENV['USERPROFILE']
+        end
+        paths << Dir.home
```
btm: @ksubrama @jdmundrawala says this is bug.

ksubrama:
This adds the "find the path to the home directory sensibly" helper code that was added to core chef to this plugin.  We copy-paste the code to allow it to work with Chef 11.  There was only one place where env home is being referenced in this plugin and I changed it to use the new helper instead.

lib/knife-windows/path_helper.rb
ksubrama: Fixed.

PR #214, opened 2015-03-27
Updated code for setting winrm port default option

Vasu1105:
:default setting for --winrm-port option was removed on master and its causing to set winrm_port to nil in knife ec2 server create. Updated code to set default option of --winrm-port to 5985 in option and if --winrm-transport value is ssl then to 5986. And removed set default part from winrm_knife_base.rb

### DIFF
--- a/lib/chef/knife/winrm_base.rb
+++ b/lib/chef/knife/winrm_base.rb
@@ -56,7 +56,8 @@ class Chef
             :long => "--winrm-transport TRANSPORT",
             :description => "The WinRM transport type.  valid choices are [ssl, plaintext]",
             :default => 'plaintext',
-            :proc => Proc.new { |transport| Chef::Config[:knife][:winrm_port] = '5986' if transport == 'ssl'; Chef::Config[:knife][:winrm_transport] = transport }
+            :proc => Proc.new { |transport| Chef::Config[:knife][:winrm_port] = '5986' if transport == 'ssl'
+                                Chef::Config[:knife][:winrm_transport] = transport }
 
           option :winrm_port,
             :short => "-p PORT",
--- a/lib/chef/knife/winrm_base.rb
+++ b/lib/chef/knife/winrm_base.rb
@@ -51,23 +51,24 @@ class Chef
             :description => "The WinRM password",
             :proc => Proc.new { |key| Chef::Config[:knife][:winrm_password] = key }
 
+          option :winrm_transport,
+            :short => "-t TRANSPORT",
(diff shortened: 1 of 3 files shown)


NimishaS:
@btm, please review this

adamedx:
The appveyor failure above was a failure to reach github, and this occurred during the well-publicized github ddos attack: http://techcrunch.com/2015/03/30/github-continues-to-face-evolving-ddos-attack/. So let's find a way to get it to re-run.

btm:
I finally figured out how to restart appveyor jobs, but now we're seeing an unrelated looking issue on the Ruby 2.0.0 test:

```
C:/Ruby200/lib/ruby/gems/2.0.0/gems/win32-api-1.4.8-x86-mingw32/lib/win32/api.rb:2:in `require': incompatible library version - C:/Ruby200/lib/ruby/gems/2.0.0/gems/win32-api-1.4.8-x86-mingw32/lib/win32/ruby19/win32/api.so (LoadError)
```

Vasu1105:
@btm I re-run the test cases and it go through. Please review.

lib/chef/knife/winrm_base.rb
```diff
@@ -51,10 +51,18 @@ def self.included(includer)
             :description => "The WinRM password",
             :proc => Proc.new { |key| Chef::Config[:knife][:winrm_password] = key }
 
+          option :winrm_transport,
+            :short => "-t TRANSPORT",
+            :long => "--winrm-transport TRANSPORT",
+            :description => "The WinRM transport type.  valid choices are [ssl, plaintext]",
+            :default => 'plaintext',
+            :proc => Proc.new { |transport| Chef::Config[:knife][:winrm_port] = '5986' if transport == 'ssl'; Chef::Config[:knife][:winrm_transport] = transport }
```
btm: Make this proc a few lines instead of using a semicolon.

btm:
The place in the unit test that you've added the port is a little irrelevant. There's a test right above it where we also test ssl but don't check the port. Its would be nice to be consistent here.

But ultimately our unit tests never really test the mixlib-cli code paths, and we know that we need integration tests if we want to test that. But we haven't built out that testing framework consistently.

So please fix the line that I commented about making multiple lines, consider making these two ssl tests consistent, then I'm :+1:.

lib/chef/knife/winrm_base.rb
Vasu1105: @btm - Done

btm:
:+1:

PR #223, opened 2015-05-04
Updated knife bootstrap windows option for consistency

Vasu1105:
1. Updated  --template-file, --distro for deprecation message to be consistent with Chef
2.  Added --bootstrap-template option to be consistent with Chef
3. --bootstrap-template short option -t is conflicting with --winrm-transport short option -t so not defining short option for --bootstrap-template at this time. We need to work on this.

### DIFF
--- a/lib/chef/knife/bootstrap_windows_base.rb
+++ b/lib/chef/knife/bootstrap_windows_base.rb
@@ -145,7 +145,7 @@ class Chef
       end
 
       def default_bootstrap_template
-        "chef-full"
+        "windows-chef-client-msi"
       end
 
       def bootstrap_template
--- a/lib/chef/knife/bootstrap_windows_base.rb
+++ b/lib/chef/knife/bootstrap_windows_base.rb
@@ -20,6 +20,8 @@ require 'chef/knife'
 require 'chef/knife/bootstrap'
 require 'chef/encrypted_data_bag_item'
 require 'chef/knife/core/windows_bootstrap_context'
+# Chef 11 PathHelper doesn't have #home
+#require 'chef/util/path_helper'
 
 class Chef
   class Knife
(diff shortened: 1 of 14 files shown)


lib/chef/knife/bootstrap_windows_base.rb
```diff
@@ -127,20 +139,37 @@ def self.included(includer)
         end
       end
 
-      # TODO: This should go away when CHEF-2193 is fixed
+      def default_bootstrap_template
+        "chef-full"
```
siddheshwar-more: It should be "windows-chef-client-msi" in windows case.

Vasu1105: Fixed

btm:
Can we avoid the code duplication with core here? Most of load_template looks the same as find_template in core. Besides the methods being organized separated slightly differently, they look the same except for newer encrypted data bag support in core.

```
  def render_template(template=nil)
    if template
      Chef::Log.warn("Passing template to render_template is deprecated in preparation of moving knife-windows into core")
    else
      template = find_template
    end

    if config[:secret_file]
      config[:secret] = Chef::EncryptedDataBagItem.load_secret(config[:secret_file])
    end

    template_content = IO.read(template).chomp
    context = Knife::Core::WindowsBootstrapContext.new(config, config[:run_list], Chef::Config)
    Erubis::Eruby.new(template_content).evaluate(context)
  end   
```

Then inherit to get #bootstrap_template and #find_template, continue to override #default_bootstrap_template?

Vasu1105:
@btm Yes duplication can be avoided. But currently I have done the changes considering current chef version that knife-window master branch is using i.e. Chef 11.

btm:
@Vasu1105 yeah :(. I'm really looking forward to getting knife-windows into core. :+1: for merge.

Issue #222, opened 2015-05-01
Validatorless bootstrap support


btm:
https://www.chef.io/blog/2015/04/16/validatorless-bootstraps/

```knife windows bootstrap winrm``` should support bootstrapping clients without a validation key like chef core does.

btm:
Fixed by #224, #234.

PR #224, opened 2015-05-07
Added validator less bootstrap support for knife bootstrap windows winrm

Vasu1105:
1) This change needs chef >= 12.2
2) --node-name option needs to be passed while bootstrap
3) Specs pending

### DIFF
--- a/lib/chef/knife/bootstrap_windows_winrm.rb
+++ b/lib/chef/knife/bootstrap_windows_winrm.rb
@@ -19,6 +19,8 @@
 require 'chef/knife/bootstrap_windows_base'
 require 'chef/knife/winrm'
 require 'chef/knife/winrm_base'
+require 'chef/knife/winrm_knife_base'
+
 
 class Chef
   class Knife
@@ -26,6 +28,8 @@ class Chef
 
       include Chef::Knife::BootstrapWindowsBase
       include Chef::Knife::WinrmBase
+      include Chef::Knife::WinrmCommandSharedFunctions
+
 
       deps do
         require 'chef/knife/core/windows_bootstrap_context'
@@ -37,6 +41,17 @@ class Chef
       banner "knife bootstrap windows winrm FQDN (options)"
 
       def run
+        if (Chef::Config[:validation_key] && !File.exist?(File.expand_path(Chef::Config[:validation_key])))
+
+          if !negotiate_auth? && !(locate_config_value(:winrm_transport) == 'ssl')
+            ui.error("Validatorless bootstrap only supported with negotiate authentication protocol and ssl/plaintext transport")
+            exit 1
+          elsif !(Chef::Platform.windows?) && negotiate_auth?
+            ui.error("Negotiate protocol with plaintext transport is only supported when this tool is invoked from windows based system")
+            exit 1
+          end
+
+        end
         bootstrap
       end
 
--- a/lib/chef/knife/bootstrap_windows_winrm.rb
+++ b/lib/chef/knife/bootstrap_windows_winrm.rb
@@ -114,4 +114,3 @@ class Chef
     end
   end
 end
-
--- a/spec/functional/bootstrap_download_spec.rb
+++ b/spec/functional/bootstrap_download_spec.rb
@@ -118,10 +118,11 @@ describe 'Knife::Windows::Core msi download functionality for knife Windows winr
     clean_test_case
 
     winrm_bootstrapper = Chef::Knife::BootstrapWindowsWinrm.new([ "127.0.0.1" ])
-    winrm_bootstrapper.define_singleton_method(:client_builder){nil}
-
-    allow(winrm_bootstrapper.client_builder).to receive(:run)
-    allow(winrm_bootstrapper.client_builder).to receive(:client_path)
+    if chef_12?
+      winrm_bootstrapper.client_builder = instance_double("Chef::Knife::Bootstrap::ClientBuilder", :run => nil, :client_path => nil)
+      allow(winrm_bootstrapper.client_builder).to receive(:run)
+      allow(winrm_bootstrapper.client_builder).to receive(:client_path)
+    end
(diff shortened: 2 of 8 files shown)


btm:
@Vasu1105, have you looked at the failing tests?

Fixes #222

Vasu1105:
@btm Yes I fixed test cases for Chef >= 12.2. The test cases failing for chef 11 and I am currently  looking into to fix those.

btm:
#### This change needs chef >= 12.2

Do you mean you need Chef >= 12.2 for this feature to work?

I presume you don't mean that this breaks knife-windows for anyone on Chef < 12.2?

#### --node-name option needs to be passed while bootstrap

Please incorporate https://github.com/chef/chef/pull/3325 to require a node be set when using this option, otherwise you see confusing output like this:

```
Node  exists, overwrite it? (Y/N) n
You said no, so I'm done here.
```

#### Specs pending

It looks like you've fixed these.

spec/functional/bootstrap_download_spec.rb
```diff
@@ -115,6 +117,10 @@ def run_download_scenario
     clean_test_case
 
     winrm_bootstrapper = Chef::Knife::BootstrapWindowsWinrm.new([ "127.0.0.1" ])
+    winrm_bootstrapper.define_singleton_method(:client_builder){nil}
+
+    allow(winrm_bootstrapper.client_builder).to receive(:run)
+    allow(winrm_bootstrapper.client_builder).to receive(:client_path)
```
btm: we need a note in here about how this is because Chef 11 doesn't have Chef::Knife::Bootstrap.client_builder.

If we're running against chef 12 (reverting 7aab00da90ad928a5a4cb123b4a300710f7e9005), we can just say:
```
winrm_bootstrapper.client_builder = instance_double("Chef::Knife::Bootstrap::ClientBuilder", :run => nil, :client_path => nil)
```

I think we need to modify travis to test both chef11 + chef12. I'm looking at that.

Vasu1105:
@btm 
1) Yes it wont break the knife-windows if chef version is < 12.2. 
2) I will be working on require --node-name change.  
3) winrm_bootstrapper.client_builder = instance_double("Chef::Knife::Bootstrap::ClientBuilder", :run => nil, :client_path => nil)  is possible  by using rspec filter where we need to add filter for chef 12 and chef 11
4) We realized that validatorless bootstrap should not be allowed if --windows-authentication-protocol set to basic. So, we are thinking to add validation for windows-authentication-protocol  and transport =ssl as well

btm:
@Vasu1105 can you rebase this against master so you pull in #229?

Then we could add some rspec filters similar to what's in Chef

```
def chef_11?
  Chef::VERSION.split('.').first.to_i == 11
end

def chef_12?
  Chef::VERSION.split('.').first.to_i == 12
end
```

Then I think we need to add code that raises an error if you don't have a validation.pem and you've got Chef 11, and a corresponding test.

Vasu1105:
@btm I am looking into all review comments and fixing them. I have rebased my branch.

lib/chef/knife/bootstrap_windows_winrm.rb
```diff
@@ -37,6 +41,17 @@ class BootstrapWindowsWinrm < Bootstrap
       banner "knife bootstrap windows winrm FQDN (options)"
 
       def run
+        if (Chef::Config[:validation_key] && !File.exist?(File.expand_path(Chef::Config[:validation_key])))
+
+          if !negotiate_auth? && !(locate_config_value(:winrm_transport) == 'ssl')
+            ui.error("Validatorless bootstrap only supported with negotiate authentication protocol and ssl/plaintext transport")
+            exit 1
+          elsif !(Chef::Platform.windows?) && negotiate_auth?
+            ui.error("Negotiate protocol with plaintext transport is only supported when this tool is invoked from windows based system")
```
btm: Doesn't this duplicate the same error being raised in `Chef::Knife::WinrmCommandSharedFunctions.resolve_winrm_transport_options` (in winrm_knife_base.rb)?

lib/chef/knife/bootstrap_windows_winrm.rb
```diff
@@ -37,6 +41,17 @@ class BootstrapWindowsWinrm < Bootstrap
       banner "knife bootstrap windows winrm FQDN (options)"
 
       def run
+        if (Chef::Config[:validation_key] && !File.exist?(File.expand_path(Chef::Config[:validation_key])))
+
+          if !negotiate_auth? && !(locate_config_value(:winrm_transport) == 'ssl')
+            ui.error("Validatorless bootstrap only supported with negotiate authentication protocol and ssl/plaintext transport")
```
btm: I still don't understand this error. Why is it not supported?

btm:
It'd be great if you could rebase this against master instead of merging and force push back onto this branch. I'm seeing some other merge commits leak in from the merge.

lib/chef/knife/bootstrap_windows_winrm.rb
smurawski: My guess is that without SSL or negotiate being used, the payload would be sent unencrypted over the wire - which would include the user.pem.  

However, this is a false assumption, since Negotiate should work with plaintext transport without encryption as well (as it does with  the winrs command).  

That assumption (which winrm-s currently makes) is why we get an error with negotiate when the remote node has allowunencrypted enabled.

btm: If you weren't using validatorless bootstrap, you'd still be sending the validator.pem over the wire. So this is a protocol problem, not a validatorless bootstrap problem.

adamedx: The issue with the validator is that this is current functionality we can't break -- we can't raise an exception when you do this. This is new, so we can make this painful.

That said, perhaps the exception is the wrong approach, maybe should just do a really scary warning to discourage this and make it very clear this should not be done except for testing purposes.

btm:
Merged via #234

PR #228, opened 2015-05-15
make winrm-s dep more strict on knife-windows 0.8.x

btm:
Previously we pinned to ~> 0.2, which would have pulled in winrm-s 0.3.0.
winrm-s 0.3.0 depends on winrm 1.3.0, which has some major changes.

Since knife-windows 1.0.0 is a major change, including the removal of em-winrm,
I want to stabilize the knife-windows 0.8.x branch.


smurawski:
:+1:

tyler-ball:
:+1:

jdmundrawala:
:+1:

PR #229, opened 2015-05-18
Update Travis/Appveyor to test Chef 11 + Chef 12

btm:
We're adding code to knife-windows to support features like validatorless bootstrap (#224) that only works on Chef 12, so we need to be able to test with Chef 12 so we can ensure we raise errors or actually work as appropriate.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 *.gem
 .bundle
 Gemfile.lock
+ci.gemfile.lock
 pkg/*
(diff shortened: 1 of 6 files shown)


btm:
@chef/client-windows, review? /cc @adamedx 

The one appveyor failure is against Chef master and will need to be looked into, but I'd like to know if everyone's happy with the other changes.

```
NoMethodError: undefined method `size' for nil:NilClass
An error occurred while installing nokogiri (1.6.6.2), and Bundler cannot
continue.
Make sure that `gem install nokogiri -v '1.6.6.2'` succeeds before bundling.
Command exited with code 5
```

ksubrama:
:+1:  I like it.

PR #230, opened 2015-05-20
run bundle install 3 times

btm:
this probably fixes appveyor in master by caching gems? It went green at least.

@jdmundrawala? /cc @ksubrama @smurawski

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -36,7 +36,7 @@ install:
   - bundler --version
 
 build_script:
-  - bundle install
+  - bundle install || bundle install || bundle install
 
 test_script:
   - bundle exec rake spec


btm:
This is an example of the failure. It happened multiple times at multiple points in the bundle install: https://ci.appveyor.com/project/Chef/knife-windows/branch/master/job/ww1pjyt4h4eyhvgh

jdmundrawala:
it's probably better the bundle vendor or whatever bundler's way of doing that is, and cache that. But i'm not quite sure I understand why this fixes things

btm:
@jdmundrawala I added the bundle install x3 pattern from core too. It passed again. Please hit merge if you think this is right.

jdmundrawala:
you should get rid of the caching if it's not needed anymore

btm:
it's hard to tell. appveyor smells funny today.

PR #232, opened 2015-05-21
Adding --hint option to bootstrap 

deepsihag:
Adding --hint option to 'knife bootstrap windows" command  in order work
with windows cloud VM instances (like ec2) and to get cloud specific
ohai attributes (like public ipaddress) and save those to node object on
chef.


btm:
Looks like a straight copy from `lib/chef/knife/bootstrap.rb`, :+1:

PR #233, opened 2015-05-21
Cache Rubygems 2.0.0 files

btm:


btm:
@jdmundrawala cache back here, at least for the majority of the builds.

Last time I was plugging ruby_version in like we do in the environment variables, but I don't know that it was really working that way, plus guessing at/switching on the API version is annoying.

I don't see any improvement here though. The Ruby 2.0.0 took longer than the 1.9.3 build anyway.

```
Environment: ruby_version=193, chef_version=< 12: 2 min 37 sec
Environment: ruby_version=200, chef_version=< 12: 2 min 47 sec
Environment: ruby_version=200, chef_version=~> 12.0: 2 min 55 sec
Environment: ruby_version=200, chef_version=master: 3 min 7 sec
```

PR #238, opened 2015-05-27
Disable caching in appveyor

jdmundrawala:
This is breaking bundler when it has to do a git checkout


btm:
:+1:  I was looking at that yesterday too.

PR #235, opened 2015-05-26
Validation key can be false

jdmundrawala:
This fixes #227.

In `boostrap_context.rb`, validation_key is defined as

```ruby
def validation_key
  if File.exist?(File.expand_path(@chef_config[:validation_key]))
    IO.read(File.expand_path(@chef_config[:validation_key]))
  else
    false
  end
end
```

PR #239, opened 2015-05-28
Fix null validation_key error

btm:
Replaces #235.

* Rebased against master with the new validatorless bootstrap support
* Updated the platform helpers in the spec tests to match Chef's bikeshed


jdmundrawala:
:+1:

PR #241, opened 2015-05-29
Knife bootstrap flag comparison test.

ksubrama:
https://github.com/chef/oc_post_mortems/blob/master/2015/03/knife-windows-chef-12.md#corrective-actions

### DIFF
--- a/lib/chef/knife/bootstrap_windows_ssh.rb
+++ b/lib/chef/knife/bootstrap_windows_ssh.rb
@@ -65,8 +65,8 @@ class Chef
         :long => "--identity-file IDENTITY_FILE",
         :description => "The SSH identity file used for authentication"
 
-      option :host_key_verification,
-        :long => "--[no-]host-key-verification",
+      option :host_key_verify,
+        :long => "--[no-]host-key-verify",
         :description => "Disable host key verification",
         :boolean => true,
         :default => true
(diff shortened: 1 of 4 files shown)


ksubrama:
https://github.com/chef/chef/pull/3469 should fix the test fail against master.

lib/chef/knife/bootstrap_windows_ssh.rb
```diff
@@ -60,14 +59,20 @@ class BootstrapWindowsSsh < Bootstrap
...
+
       option :identity_file,
         :short => "-i IDENTITY_FILE",
         :long => "--identity-file IDENTITY_FILE",
         :description => "The SSH identity file used for authentication"
 
-      option :host_key_verification,
-        :long => "--[no-]host-key-verification",
-        :description => "Disable host key verification",
+      option :host_key_verify,
```
btm: If we change this we probably need some kind of backwards-compat / deprecation message?

spec/unit/knife/bootstrap_options_spec.rb
```diff
@@ -0,0 +1,145 @@
...
+    let(:opt_map) { {
+      :msi_url => :bootstrap_url,
+      :encrypted_data_bag_secret => :secret,
+      :encrypted_data_bag_secret_file => :secret_file,
+      :winrm_user => :ssh_user,
+      :winrm_password => :ssh_password,
+      :winrm_port => :ssh_port,
+      :winrm_ssl_verify_mode => :host_key_verify,
+    }}
+    let(:ref_ignore) { [
```
btm: we should separate this into two arrays, one of things that we know are different and it's okay, and one that is a punch list of stuff we need to fix. and put a comment on them explaining that.

jdmundrawala:
LGTM :+1:

PR #246, opened 2015-06-04
Update winrm error message to provide correct options

btm:
The error message regarding negotiate when on non-windows platforms
specifies an incorrect option alternative. This commit improves that
error message.


ksubrama:
:+1:

Issue #244, opened 2015-06-04
winrm bootstrap failures don't fail until they timeout


btm:
When knife-windows has serious error trying to bootstrap, we should fail immediately.:

Example 1: https://github.com/chef/knife-windows/issues/210 / https://github.com/chef/knife-windows/issues/218

```
Waiting for remote response before bootstrap.......................ERROR: No response received from remote node after 1.86 minutes, giving up.
ERROR: RuntimeError: Error: Unencrypted communication not supported. Please check winrm configuration winrm/config/service AllowUnencrypted flag.
```

Example 2: When using --winrm-authentication basic from Linux but AllowUnencrypted is False:

```
$ knife bootstrap windows winrm 192.168.5.112 -x Administrator -P 'P@ssw0rd' --winrm-authentication-protocol basic

Waiting for remote response before bootstrap.......................ERROR: No response received from remote node after 1.84 minutes, giving up.
ERROR: WinRM::WinRMAuthorizationError: WinRM::WinRMAuthorizationError
```

(We should give a useful suggestion on that one too)

btm:
Fixed by #252.

PR #252, opened 2015-06-15
Fixed winrm bootstrap failures don't fail until they timeout 

siddheshwar-more:
Issue ref- https://github.com/chef/knife-windows/issues/244

### DIFF
--- a/lib/chef/knife/bootstrap_windows_winrm.rb
+++ b/lib/chef/knife/bootstrap_windows_winrm.rb
@@ -89,18 +89,14 @@ class Chef
           raise RuntimeError, 'Command execution failed.' if status != 0
           ui.info(ui.color("Remote node responded after #{elapsed_time_in_minutes(wait_start_time)} minutes.", :magenta))
           return
-        rescue
+        rescue Errno::ECONNREFUSED => e
+          ui.error("Connection refused connecting to #{locate_config_value(:server_name)}:#{locate_config_value(:winrm_port)}.")
+          raise
+        rescue Exception => e
           retries_left -= 1
-          # $! will contain the current exception object
-          if $!.class.equal?(Errno::ECONNREFUSED)
-            ui.error("RuntimeError: No response received from remote node.")
-            ui.info("Hint: Please check remote server's winrm configuration i.e '$ winrm/config/service' AllowUnencrypted flag")
-            raise
-          end
-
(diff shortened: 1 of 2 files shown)


lib/chef/knife/bootstrap_windows_winrm.rb
```diff
@@ -91,8 +91,16 @@ def wait_for_remote_response(wait_max_minutes)
           return
         rescue
           retries_left -= 1
+          # $! will contain the current exception object
+          if $!.class.equal?(Errno::ECONNREFUSED)
```
adamedx: Shouldn't this be `rescue Errno::ECONFFREFUSED => e` to avoid explicitly checking the type of the exception instance?

adamedx:
Should probably have a unit spec for this as documentation and also to prevent regression.

lib/chef/knife/bootstrap_windows_winrm.rb
btm: Oh, I didn't realize we were doing a blind rescue.

* We should print the exception in the error message on the blind rescue to facilitate doing the right thing when we're rescuing something we shouldn't
* I'm not sure that AllowUnencrypted would cause a Errno::ECONNREFUSED. I would leave out the hint about checking AllowUnencrypted for the Errno::ECONNREFUSED case.

Yeah, this should be something like:

```
  rescue Errno::ECONNREFUSED => e
    ui.error("Connection refused connecting to HOST:PORT")
    raise
  rescue Exception => e  
    # timeout stuff
    ui.error("No response received from remote node after #{elapsed_time_in_minutes(wait_start_time)} minutes, giving up.")
    ui.error("Exception: #{e.message}")
    raise
  end
```

btm:
I agree with @adamedx's comments.

lib/chef/knife/bootstrap_windows_winrm.rb
siddheshwar-more: @adamedx  @btm 
Fixed review comments. Thanks!

NimishaS:
@btm, please review if this PR is suitable for merging.

btm:
Rebased and merged.

I'm a little concerned that there might be a valid use case where you would get ECONNREFUSED a couple times, but not sure. We will wait and see.

PR #260, opened 2015-07-01
Negotiate warning for non-windows platforms is late

Vasu1105:


Vasu1105:
@btm Handling this in mixlib-cli proc block is not possible since proc is not executed in case of default value for option. I right now added the validation in run method of bootstrap_windows_winrm.

PR #268, opened 2015-07-23
Use winrm 1.3.x in 0.8-stable

jaym:


Gemfile
```diff
@@ -3,6 +3,7 @@ source "https://rubygems.org"
 # Specify your gem's dependencies in knife-windows.gemspec
 gemspec
 
+gem 'em-winrm', git: 'https://github.com/chef/em-winrm'
```
adamedx: This needs to go, right?

adamedx:
Other than the git dep in the gemfile, I'm good on this one.

Gemfile
jaym: yep

ksubrama:
:+1:

PR #272, opened 2015-07-30
Fix for secret-file not being rendered in knife windows.

Aliasgar16:
Code changes testing is in Progress for this PR.

### DIFF
--- a/lib/chef/knife/bootstrap_windows_base.rb
+++ b/lib/chef/knife/bootstrap_windows_base.rb
@@ -225,7 +225,7 @@ class Chef
 
       def render_template(template=nil)
         if config[:secret_file]
-          config[:secret] = Chef::EncryptedDataBagItem.load_secret(config[:secret_file])
+          config[:secret] = Chef::EncryptedDataBagItem.load_secret(config[:secret_file]) if !config[:secret]
         end
         Erubis::Eruby.new(template).evaluate(bootstrap_context)
       end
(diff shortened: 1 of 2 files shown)


lib/chef/knife/bootstrap_windows_base.rb
```diff
@@ -236,6 +236,10 @@ def bootstrap(proto=nil)
           config[:secret_file] ||= Chef::Config[:knife][:encrypted_data_bag_secret_file]
           config[:secret] ||= Chef::Config[:knife][:encrypted_data_bag_secret]
         end
+        if config[:encrypted_data_bag_secret_file] || config[:encrypted_data_bag_secret]
```
siddheshwar-more: Can we use `locate_config_value(:encrypted_data_bag_secret_file)` ?

lib/chef/knife/bootstrap_windows_base.rb
```diff
@@ -236,6 +236,10 @@ def bootstrap(proto=nil)
...
+          config[:secret_file] = config[:encrypted_data_bag_secret_file]
+          config[:secret] = config[:encrypted_data_bag_secret] || Chef::EncryptedDataBagItem.load_secret(config[:secret_file]) if config[:encrypted_data_bag_secret] || config[:secret_file]
+        end
```
siddheshwar-more: Please add rspec.

spec/unit/knife/bootstrap_windows_winrm_spec.rb
```diff
@@ -41,6 +42,109 @@
 
   let(:initial_fail_count) { 4 }
 
+  context "knife secret-file && knife secret options are passed" do
```
adamedx: It might be better to move this test under functional rather than unit since it makes changes to the file system. If everything can be fully mocked out, then keeping it a unit test is fine.

Aliasgar16: I have modified the RSpecs to mock out the behavior instead of creating secret-file on the file system, keeping the specs under unit test cases.

NimishaS:
@btm please review this.

ksubrama:
@btm is on leave :)  This looks good to me. :ship:

PR #273, opened 2015-08-10
Update bootstrap test to include msi_url with space

chefsalim:
Adding a test to check whether the msi_url flag can handle a path with a space.

@adamedx @schisamo @ksubrama

### DIFF
--- a/spec/functional/bootstrap_download_spec.rb
+++ b/spec/functional/bootstrap_download_spec.rb
@@ -100,6 +100,14 @@ describe 'Knife::Windows::Core msi download functionality for knife Windows winr
         run_download_scenario
       end
     end
+
+     context "when provided a custom msi_url with space in path to fetch from" do
+      let(:mock_bootstrap_context) { Chef::Knife::Core::WindowsBootstrapContext.new(
+        { :msi_url => "file:///C:/Program Files/Windows NT/Accessories/wordpad.exe" }, nil, { :knife => {} }) }
+      it "downloads the chef-client MSI from a custom path with spaces during winrm bootstrap" do
+        run_download_scenario
+      end
+    end
   end
 
   def download_succeeded?


spec/functional/bootstrap_download_spec.rb
```diff
@@ -100,6 +100,14 @@
         run_download_scenario
       end
     end
+
+     context "when provided a custom msi_url with space in path to fetch from" do
+      let(:mock_bootstrap_context) { Chef::Knife::Core::WindowsBootstrapContext.new(
+        { :msi_url => "file:///C:/Program Files/Windows NT/Accessories/wordpad.exe" }, nil, { :knife => {} }) }
```
adamedx: So does this actually download the file? If it does that will make things significantly slower...

ksubrama: It's copying from the local system and this is a functional test.  It should be fast enough.  We copy xcopy.exe above and that's been quite fast.

spec/functional/bootstrap_download_spec.rb
```diff
@@ -100,6 +100,14 @@
...
+      it "downloads the chef-client MSI from a custom path with spaces during winrm bootstrap" do
```
ksubrama: Can you add a test for the other direction of this - does it still work if I do file://C:/Program%20Files ... ?

chefsalim:
@ksubrama No, the test does not work if the msi_url has a space escaped to a %20 already, since % is an escape character for the Windows cmd line interpreter itself.  The following generated code is used to set up the Powershell wget command:

@set powershell_download=powershell.exe -ExecutionPolicy Unrestricted -NoProfile -NonInteractive -File  C:\Users\salim\AppData\Local\Temp\bootstrap_test20150811-1380-1wc8u44\wget.ps1 "%REMOTE_SOURCE_MSI_URL%" "%LOCAL_DESTINATION_MSI_PATH%"

ksubrama:
:+1:

ksubrama:
After discussing this with @adamedx , there are a few yaks to correctly support % in a url as part of a bootstrap script.  We're punting on supporting it.  This is good as it is.

PR #275, opened 2015-09-02
Added bootstrap_install_command option in parity with knife bootstrap…

Aliasgar16:
… command and made necessary changes in RSpecs for the same.

### DIFF
--- a/lib/chef/knife/bootstrap/windows-chef-client-msi.erb
+++ b/lib/chef/knife/bootstrap/windows-chef-client-msi.erb
@@ -105,85 +105,91 @@ goto install
 goto install
 
 :install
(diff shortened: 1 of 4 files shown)


NimishaS:
@btm , @ksubrama please review this.

adamedx:
This looks good, but it needs tests and some parameter validation. I think a unit test in the existing bootstrap context spec, and a functional test here::https://github.com/chef/knife-windows/blob/master/spec/functional/bootstrap_download_spec.rb. You'd just test to see that if you set the install-command config, it actually renders it in the template, possibly by actually executing the rendered content on Windows (which would also ensure that there are no syntax errors).

PR #276, opened 2015-09-03
Improve error handling for non basic authentication

mwrock:
The current authentication failure messaging is specific to basic authentication and provides advise for turning off encryption on the target node. Failures for Negotiate authentication are not caught and just surface the `WinRMAuthorizationError`. This PR specifically catches `WinRMAuthorizationError` and provides a different Hint to the user when not using Basic auth.

### DIFF
--- a/lib/chef/knife/winrm.rb
+++ b/lib/chef/knife/winrm.rb
@@ -29,6 +29,13 @@ class Chef
 
       include Chef::Knife::WinrmCommandSharedFunctions     
 
+      FAILED_BASIC_HINT ||= "Hint: Please check winrm configuration 'winrm get winrm/config/service' AllowUnencrypted flag on remote server."
+      FAILED_NOT_BASIC_HINT ||= <<-eos.gsub /^\s+/, ""
+        Hint: Make sure to prefix domain usernames with the correct domain name.
+        Hint: Local user names should be prefixed with computer name or IP address.
+        EXAMPLE: my_domain\\user_namer
+      eos
+
       deps do
         require 'readline'
         require 'chef/search/query'
@@ -69,14 +76,13 @@ class Chef
             exit @exit_code if @exit_code && @exit_code != 0
             @exit_code || 0
           end
-        rescue WinRM::WinRMHTTPTransportError => e
-          case e.message
-          when /401/
+        rescue WinRM::WinRMHTTPTransportError, WinRM::WinRMAuthorizationError => e
+          if authorization_error?(e)
             if ! config[:suppress_auth_failure]
               # Display errors if the caller hasn't opted to retry
               ui.error "Failed to authenticate to #{@name_args[0].split(" ")} as #{locate_config_value(:winrm_user)}"
               ui.info "Response: #{e.message}"
-              ui.info "Hint: Please check winrm configuration 'winrm get winrm/config/service' AllowUnencrypted flag on remote server."
+              ui.info get_failed_authentication_hint
               raise e
             end
             @exit_code = 401
@@ -180,11 +186,24 @@ class Chef
         Readline
       end
 
+      def authorization_error?(exception)
+        exception.is_a?(WinRM::WinRMAuthorizationError) ||
+          exception.message =~ /401/
+      end
+
       def success_return_codes
         #Redundant if the CLI options parsing occurs
         return [0] unless config[:returns]
         return @success_return_codes ||= config[:returns].split(',').collect {|item| item.to_i}
-      end      
+      end
+
+      def get_failed_authentication_hint
+        if @session_opts[:basic_auth_only]
+          FAILED_BASIC_HINT
+        else
+          FAILED_NOT_BASIC_HINT
+        end
+      end
     end
   end
 end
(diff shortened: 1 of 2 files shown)


mwrock:
Odd, the tests pass on my box and i am using `bundle exec`. Will investigate...may be a windows thing.

Rakefile
```diff
@@ -4,11 +4,16 @@ Bundler::GemHelper.install_tasks
 begin
   require 'rspec/core/rake_task'
 
-  task :default => :spec
+  task :default => [:unit_spec, :functional_spec]
 
   desc "Run all specs in spec directory"
```
ksubrama: Fix descriptions

spec/unit/knife/winrm_spec.rb
```diff
@@ -272,9 +272,10 @@
           expect(exit_code).to be_zero
         end
 
-        it "exits with 100 if command execution raises an exception other than 401" do
+        it "exits with 100 and nom hint if command execution raises an exception other than 401" do
```
ksubrama: typo "no hint"

ksubrama:
Check appveyor.yml  It's doing bundle exec rake spec  travis.yml doesn't specify anything.  I think your rake file cleanup may be causing them to run different tests (maybe?).

mwrock:
Heh. Just took a closer look at the Travis output. It's because the build runs on Linux and I ran the tests in windows. Just need to change the test to fake the windows platform.

ksubrama:
D'oh - I thought the unit test helpers already did that in knife-windows.  I guess not.

mwrock:
squashed things up and working to fix new build breaks as a result of `tags` being added to core bootstrap. Will push that fix ini a separate commit.

smurawski:
:+1:

Rakefile
```diff
@@ -4,11 +4,16 @@ Bundler::GemHelper.install_tasks
 begin
   require 'rspec/core/rake_task'
 
-  task :default => :spec
+  task :default => [:unit_spec, :functional_spec]
```
mcquin: I'm for this change, but against making it in this PR. It's not within scope.

mwrock: thats cool. I kept it in a separate commit and can merge separately

chefsalim: Is it basically a no-op or does it actually make a difference?

mwrock: @chefsalim not sure i get the question, but it just separates functional and unit tests into separate tasks but the default task (what ci uses) will still run both.

smurawski: @chefsalim it doesn't break anything but improves the dev experience.

lib/chef/knife/bootstrap_windows_base.rb
```diff
@@ -170,6 +170,12 @@ def self.included(includer)
               Chef::Config[:knife][:bootstrap_vault_item][vault].push(item)
               Chef::Config[:knife][:bootstrap_vault_item]
             }
+
+          option :tags,
+            :long => "--tags TAGS",
+            :description => "Comma separated list of tags to apply to the node",
+            :proc => lambda { |o| o.split(/[\s,]+/) },
+            :default => []
```
mcquin: Why was this added, @mwrock ?

mcquin: I see now, it's for the `first_boot` change. I think this should go in a separate PR, too.

mwrock: yup. this is gonna be 3 PRs. Apologies for not doing that up front and thanks for noticing!

mwrock: It does feel like this could be "dried up" Seems like we basically have to inherit via copy/paste when new properties are added to https://github.com/chef/chef/blob/master/lib/chef/knife/bootstrap.rb

would be nice if we could add some DSL like:
```
include_option :runlist
include_option :tags
```

That would also mean not having to do the crazy versioning filters in the tests too

lib/chef/knife/winrm.rb
```diff
@@ -29,6 +29,13 @@ class Winrm < Knife
 
       include Chef::Knife::WinrmCommandSharedFunctions     
 
+      FAILED_BASIC_HINT ||= "Hint: Please check winrm configuration 'winrm get winrm/config/service' AllowUnencrypted flag on remote server."
```
chefsalim: Reduce the line length?  It is getting truncated in the browser view.

lib/chef/knife/winrm.rb
```diff
@@ -29,6 +29,13 @@ class Winrm < Knife
...
+      FAILED_NOT_BASIC_HINT ||= <<-eos.gsub /^\s+/, ""
```
chefsalim: Left justify the text so we don't have to do a gsub?

mwrock: I personally prefer the `gsub` (though I'm not crazy about it) over losing the indentation.

Any one else have opinions here?

mcquin: I'm fine with the `gsub` as well. There are occurrences of both `gsub`bing and left-justifying all over our code bases.

mcquin:
LGTM

lib/chef/knife/winrm.rb
chefsalim: Not a big deal - I'm fine with the aestheticly pleasing way :)

mwrock: cool

PR #281, opened 2015-09-05
fixes #280 preventing unencrypted negotiate auth on non windows

### LINKED ISSUE #280: Negotiate auth without encryption warning should not be fatal
The following warning is actually treated as an error and causes the tool to exit. That is not the correct behavior -- we should simply warn and continue. Also, the message is not right -- it claims that this is not supported, however it *is* valid and actually works if you remove the "exit" from the affected source code: 

https://github.com/chef/knife-windows/blob/v1.0.0.rc.1/lib/chef/knife/winrm_knife_base.rb#L130

For this to work, the remote server must have "allowunencrypted" set to "$true" for WinRM config. This can be validated running knife from Mac OSX.

The fix here should be the following:

* Remove the exit.
* Change the warning text from this:

> Using '--winrm-authentication-protocol negotiate' with '--winrm-transport plaintext' is only supported when this tool is invoked from a Windows system. Switch to either '--winrm-transport ssl' or '--winrm-authentication-protocol basic

  to something like this:

> You are using '--winrm-authentication-protocol negotiate' with '--winrm-transport plaintext' on a non-Windows system which results in unencrypted traffic. To avoid this warning and secure communication, use '--winrm-transport ssl' instead of the plaintext transport, or execute this command from a Windows system which enables encrypted communication over plaintext with the negotiate authentication protocol.

Specs should also reflect the desired behavior.

mwrock:
this prevents unencrypted negotiate auth on non windows from failing and improves the warning message. To prevent this warning from being displayed on every winrm command during bootstrap, the winrm bootstrap command is refactored to validate this authentication method on its own and then call a single winrm session for all winrm calls instead of calling run on a new winrm knife instance for each call that would perform the validation each time.

### DIFF
--- a/lib/chef/knife/bootstrap_windows_winrm.rb
+++ b/lib/chef/knife/bootstrap_windows_winrm.rb
@@ -39,9 +39,7 @@ class Chef
 
       banner 'knife bootstrap windows winrm FQDN (options)'
 
-      def run
-        validate_options!
-
+      def run       
         if (Chef::Config[:validation_key] && !File.exist?(File.expand_path(Chef::Config[:validation_key])))
           if !negotiate_auth? && !(locate_config_value(:winrm_transport) == 'ssl')
             ui.error('Validatorless bootstrap over unsecure winrm channels could expose your key to network sniffing')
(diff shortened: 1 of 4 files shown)


mwrock:
Hmm. I'll setup an ec2 environment Tuesday to try. I've been trying on linux with a local windows node bootstrapped against the hosted service and the unencrypted bit flipped on the win node. Thanks for trying late on a Friday!

mwrock:
Just took a closer look at the winrm transport resolution and see why it would be getting set to `:kerberos` if not coming in via vanilla windows bootstrap like I have been testing with. Have to leave for the weekend but think this should be pretty simple.I definitely need to be testing this stuff against ec2 and azure.

adamedx:
You may be able to simply repro without knife-ec2 just using knife bootstrap.

mwrock:
This is what I had been using.
```
knife bootstrap windows winrm 192.168.1.14 -P vagrant -x '192.168.1.14\vagrant' -N tk1
```
I think I just pushed the fix but will be in a better place to fully test Tuesday. From what I can tell only ec2 is affected (not azure) but I will test azure too. 

Ec-2 passes the kerberos config along to knife-windows even if all the keys were unset. That should be fine, but the code I had deleted only forwarded those settings if they were set. The internal knife-windows implementation sets the transport to `:kerberos` if any kerberos key simply exists regardless of whether all values are nil. This fix adds a nil check. I added unit tests to verify the fix but will test against both azure and ec-2 Tuesday.

mwrock:
Validated on ec-2 and azure

adamedx:
It's working for me now, thank you @mwrock. :+1:

smurawski:
LGTM :+1:

jkeiser:
:+1:

PR #282, opened 2015-09-08
prefix local usernames with dot domain notation when using negotiate

mwrock:
Under some circumstances, negotiate authentication will fail using a local account if the "local domain" is not specified. Users may likely neglect to prefix a local username with the machine name or IP address and its confusing why access is being denied. This PR detects if an explicit domain is not being sent when using negotiate and if not, uses the short form `.\{user}` username.

### DIFF
--- a/lib/chef/knife/bootstrap_windows_winrm.rb
+++ b/lib/chef/knife/bootstrap_windows_winrm.rb
@@ -47,8 +47,9 @@ class Chef
           end
         end
 
-        validate_options
+        validate_options!
         resolve_session_options
+        load_windows_specific_gems if @session_opts[:transport] == :sspinegotiate
         @session_opts[:host] = server_name
         @session = Chef::Knife::WinrmSession.new(@session_opts)
 
--- a/lib/chef/knife/winrm_knife_base.rb
+++ b/lib/chef/knife/winrm_knife_base.rb
@@ -32,9 +32,15 @@ class Chef
           include Chef::Knife::WinrmBase
           include Chef::Knife::WinrmSharedOptions
 
-          #Overrides Chef::Knife#configure_session, as that code is tied to the SSH implementation
-          #Tracked by Issue # 3042 / https://github.com/chef/chef/issues/3042
-          def validate_options
+          def validate_options!
+            winrm_auth_protocol = locate_config_value(:winrm_authentication_protocol)
+
+            if ! Chef::Knife::WinrmBase::WINRM_AUTH_PROTOCOL_LIST.include?(winrm_auth_protocol)
+              ui.error "Invalid value '#{winrm_auth_protocol}' for --winrm-authentication-protocol option."
+              ui.info "Valid values are #{Chef::Knife::WinrmBase::WINRM_AUTH_PROTOCOL_LIST.join(",")}."
+              exit 1
+            end
+
             if negotiate_auth? && !Chef::Platform.windows? && !(locate_config_value(:winrm_transport) == 'ssl')
               ui.warn <<-eos.gsub /^\s+/, ""
                 You are using '--winrm-authentication-protocol negotiate' with 
@@ -47,9 +53,13 @@ class Chef
             end
           end
 
+          #Overrides Chef::Knife#configure_session, as that code is tied to the SSH implementation
+          #Tracked by Issue # 3042 / https://github.com/chef/chef/issues/3042
           def configure_session
-            validate_options
+            validate_options!
             resolve_session_options
+            load_windows_specific_gems if @session_opts[:transport] == :sspinegotiate
+            warn_no_ssl_peer_verification if @session_opts[:no_ssl_peer_verification]
             resolve_target_nodes
             session_from_list
           end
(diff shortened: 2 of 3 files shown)


mwrock:
Just moved this user name resolution out of `winrm_session` and into the shared winrm shared methods with the rest of the session option resolution logic.

This includes a refactoring of the option resolution logic. With the exception of the username the outcome of the code should be the same and thus those unit tests remain unchanged. The refactoring treats each option atomically and removes the dependencies that some options had on how another option was resolved requiring some options to be resolved before other options.

lib/chef/knife/winrm_knife_base.rb
```diff
@@ -46,6 +54,8 @@ def validate_options!
           def configure_session
             validate_options!
             resolve_session_options
+            load_windows_specific_gems if @session_opts[:transport] == :sspinegotiate
+            warn_no_ssl_peer_verification if @session_opts[:no_ssl_peer_verification]
```
mwrock: note this line forces a warning even when NOT using SSL. Rather than addressing here, I'm gonna wait for this to go through and submit separately.

jkeiser: What is the effect if you leave the warning out entirely until that fix? We've learned recently that excessive warnings can be pretty devastating to user experience.

lib/chef/knife/winrm_knife_base.rb
```diff
@@ -99,71 +109,68 @@ def create_winrm_session(options={})
...
+            kerberos_opts[:service] = locate_config_value(:kerberos_service) if locate_config_value(:kerberos_service)
+            kerberos_opts
           end
 
-          def resolve_winrm_auth_settings
-            winrm_auth_protocol = locate_config_value(:winrm_authentication_protocol)
-            if ! Chef::Knife::WinrmBase::WINRM_AUTH_PROTOCOL_LIST.include?(winrm_auth_protocol)
-              ui.error "Invalid value '#{winrm_auth_protocol}' for --winrm-authentication-protocol option."
-              ui.info "Valid values are #{Chef::Knife::WinrmBase::WINRM_AUTH_PROTOCOL_LIST.join(",")}."
-              exit 1
```
mcquin: It looks like this warning has been removed. Is this intentional?

lib/chef/knife/winrm_knife_base.rb
```diff
@@ -99,71 +109,68 @@ def create_winrm_session(options={})
...
-            @session_opts[:port] = locate_config_value(:winrm_port)
-
-            #30 min (Default) OperationTimeout for long bootstraps fix for KNIFE_WINDOWS-8
-            @session_opts[:operation_timeout] = locate_config_value(:session_timeout).to_i * 60 if locate_config_value(:session_timeout)
+          def resolve_winrm_user
+            user = locate_config_value(:winrm_user)
+            
+            if resolve_winrm_basic_auth ||
+              resolve_winrm_transport == :kerberos ||
+              user.include?("\\") ||
```
jkeiser: Are we guaranteed `locate_config_value(:winrm_user)` will have a value? I see a crash in otherwise. There are a few other spots where crashes may happen as well.

lib/chef/knife/winrm_knife_base.rb
```diff
@@ -99,71 +109,68 @@ def create_winrm_session(options={})
...
 
-          def resolve_winrm_kerberos_options
-            if config.keys.any? {|k| k.to_s =~ /kerberos/ }
-              @session_opts[:transport] = :kerberos
-              @session_opts[:keytab] = locate_config_value(:kerberos_keytab_file) if locate_config_value(:kerberos_keytab_file)
-              @session_opts[:realm] = locate_config_value(:kerberos_realm) if locate_config_value(:kerberos_realm)
-              @session_opts[:service] = locate_config_value(:kerberos_service) if locate_config_value(:kerberos_service)
-            end
+          def resolve_winrm_session_timeout
+            locate_config_value(:session_timeout).to_i * 60 if locate_config_value(:session_timeout)
```
mcquin: Can you re-add the comment that was here? (https://github.com/chef/knife-windows/pull/282/files#diff-4a7d2c3d64141072c2155c9e8428d277L115)

jkeiser:
General suggestion (not always practical): if you are refactoring and making user-impacting changes at the same time, it can help the reviewer a lot if you put the refactor into a separate commit. I know when I see too much going on I tend to look at the separate commits to see if I can parse it out better. Don't change this one though :)

lib/chef/knife/winrm_knife_base.rb
smurawski: @jkeiser There's a default for :winrm_user of administrator if it isn't resolved from the knife config or command line.

lib/chef/knife/winrm_knife_base.rb
```diff
@@ -99,71 +109,68 @@ def create_winrm_session(options={})
...
+          def resolve_winrm_user
+            user = locate_config_value(:winrm_user)
+            
+            if resolve_winrm_basic_auth ||
+              resolve_winrm_transport == :kerberos ||
+              user.include?("\\") ||
+              user.include?("@")
+              user
+            else
+              ".\\#{user}"
```
jkeiser: Could you put a comment explaining the decision being made here and why it needs a prefix? It will be helpful for people who modify this code in the future.

lib/chef/knife/winrm_knife_base.rb
jkeiser: Awesome.

lib/chef/knife/winrm_knife_base.rb
mwrock: No I moved it to `validate_options!` which is the method responsible for warning and exiting on bad options.

mwrock:
Good Point @jkeiser : totally should have added the user resolution as a separate commit.

jkeiser:
Minor comments, :+1: with answers or fixes :)

mwrock:
@jkeiser @mcquin comments added. Thanks!!

lib/chef/knife/bootstrap_windows_winrm.rb
```diff
@@ -47,8 +47,9 @@ def run
           end
         end
 
-        validate_options
+        validate_options!
         resolve_session_options
+        load_windows_specific_gems if @session_opts[:transport] == :sspinegotiate
```
mwrock: Will be submitting a separate PR moving this dependency loading into `WinrmSession` since it is what actually takes the dependency and that will remove the duplicate calls.

PR #286, opened 2015-09-09
moving winrm-s require intro winrm_session

mwrock:


mwrock:
after discussing with @adamedx I made the dependency loaderr a class method which meant I could remove some crappy code (mine).

jaym:
:+1:

smurawski:
:+1:

Issue #204, opened 2015-03-06
Default parameter values do not always resolve properly.


smurawski:
On a number of unit tests, the default parameter values do not show up in the config hash.  To work around this, I updated the locate_config_value function in winrm_base to check default_config as well.  The real source as to why these values aren't resolving properly needs to be tracked down.

PR #287, opened 2015-09-11
Removed duplicate declaration of locate_config_value method.

Aliasgar16:


lib/chef/knife/bootstrap_windows_base.rb
```diff
@@ -362,11 +362,6 @@ def bootstrap_bat_file
         @bootstrap_bat_file ||= "\"%TEMP%\\bootstrap-#{Process.pid}-#{Time.now.to_i}.bat\""
       end
 
-      def locate_config_value(key)
```
mwrock: Nice!

lib/chef/knife/winrm_base.rb
```diff
@@ -117,7 +117,6 @@ def self.included(includer)
       def locate_config_value(key)
         key = key.to_sym
         value = config[key] || Chef::Config[:knife][key] || default_config[key]
-        Chef::Log.debug("Looking for key #{key} and found value #{value}")
```
mwrock: I'd keep this as it can be helpful for troubleshooting.

Aliasgar16:
@mwrock the method `locate_config_value` has been declared at two places under `knife-windows` https://github.com/chef/knife-windows/blob/master/lib/chef/knife/winrm_base.rb#L117 and https://github.com/chef/knife-windows/blob/master/lib/chef/knife/bootstrap_windows_base.rb#L365.
You are right that removing this method from `bootstrap_windows_base` will break bootstrap ssh. We should ideally have this method defined only once.

mwrock:
Totally agree we should not have 2. In fact, we really shouldn't even have 1 :) (see #204).

lib/chef/knife/winrm_base.rb
adamedx: Yes, debug is ok, that's not the default, so we should keep this.

mwrock:
@Aliasgar16 consider creating a `KnifeWindowsBase` module that defines `locate_config_value` and would replace the other 2.

adamedx:
@mwrock, thoughts on this after the updates?

mwrock:
LGTM :+1:

PR #301, opened 2015-10-07
defer nokogiri require

mwrock:
ChefDK carries along 2 versions of nokogiri and the older one is hard pinned to mini_portile 0.6.0 so we also get 2 versions of mini_portile as well which is what eventually causes this warning. The root problem to be addressed separately is the fact that chefDK grabs 2 versions of nokogiri and based on a `gem dependency -R nokogiri` there appears to be no good reason for this and we should just get the latter.

However I see this as a worthwhile change to quicky fix the problem here and deferring gem loading is a good thing in this case.

### DIFF
--- a/lib/chef/knife/wsman_test.rb
+++ b/lib/chef/knife/wsman_test.rb
@@ -17,11 +17,9 @@
 #
 
 require 'httpclient'
-require 'nokogiri'
 require 'chef/knife'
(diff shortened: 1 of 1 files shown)


adamedx:
Interesting -- we missed the pry inclusion as well. :+1: . I'll cut a new release once this gets merged if @btm is ok with it.

btm:
:+1:

PR #302, opened 2015-10-08
Address regression caused by chef client 12.5 environment argument

adamedx:
I believe this is the commit in Chef (release in 12.5.1) that caused the regression in knife-windows: https://github.com/chef/chef/commit/14d7baab6365e0b4146990a5306f80a7f0d09a82 -- cc @danielsdeleo. I'm just not exactly sure how...

### DIFF
--- a/lib/chef/knife/core/windows_bootstrap_context.rb
+++ b/lib/chef/knife/core/windows_bootstrap_context.rb
@@ -140,9 +140,9 @@ CONFIG
         end
 
         def start_chef
-          bootstrap_environment_option = bootstrap_environment.nil? ? '' : "-E #{bootstrap_environment}"
+          bootstrap_environment_option = bootstrap_environment.nil? ? '' : " -E #{bootstrap_environment}"
           start_chef = "SET \"PATH=%PATH%;C:\\ruby\\bin;C:\\opscode\\chef\\bin;C:\\opscode\\chef\\embedded\\bin\"\n"
-          start_chef << "chef-client -c c:/chef/client.rb -j c:/chef/first-boot.json #{bootstrap_environment_option}\n"
+          start_chef << "chef-client -c c:/chef/client.rb -j c:/chef/first-boot.json#{bootstrap_environment_option}\n"
         end
 
         def latest_current_windows_chef_version_query
--- a/lib/chef/knife/core/windows_bootstrap_context.rb
+++ b/lib/chef/knife/core/windows_bootstrap_context.rb
@@ -140,8 +140,9 @@ CONFIG
         end
 
         def start_chef
+          bootstrap_environment_option = bootstrap_environment.nil? ? '' : "-E #{bootstrap_environment}"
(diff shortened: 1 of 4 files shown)


mwrock:
You may have already caught on to this, but you will need to update the support templates and remove the `-E _defaut`

adamedx:
Yeah, in the process of adding some 12.5 specific specs.

adamedx:
@mwrockx, @btm, @chefsalim can you take a look?

mwrock:
looks good but I don't think you meant to add the ~ files. :+1:  after those get deleted.

adamedx:
What `~` files? :smile: Thank you @mwrock!

mwrock:
Are you using MS Word as an editor?

adamedx:
hehe, just emacs, which will give you files with `~`, `#`, and sometimes both. Should probably update .gitignore...

mwrock:
:+1:  again.

FYI @btm @chefsalim  bootstraps are broken until this is merged. I had to manually add the fix while testing the policyfile support to get it to work. Would bwe greato to get another +1 soon

danielsdeleo:
:+1: :shipit:

lib/chef/knife/core/windows_bootstrap_context.rb
```diff
@@ -140,8 +140,9 @@ def config_content
         end
 
         def start_chef
+          bootstrap_environment_option = bootstrap_environment.nil? ? '' : " -E #{bootstrap_environment}"
           start_chef = "SET \"PATH=%PATH%;C:\\ruby\\bin;C:\\opscode\\chef\\bin;C:\\opscode\\chef\\embedded\\bin\"\n"
-          start_chef << "chef-client -c c:/chef/client.rb -j c:/chef/first-boot.json -E #{bootstrap_environment}\n"
+          start_chef << "chef-client -c c:/chef/client.rb -j c:/chef/first-boot.json#{bootstrap_environment_option}\n"
```
btm: Does mean knife-windows requires Chef 12.5+ now? If so, can we say something in the README?

mwrock: I think this is backwards compatible. Before if `-E` was ommited, we would add `-E -default` now we just omit it alltogether which I believe is effectively the same.

spec/functional/bootstrap_download_spec.rb
```diff
@@ -169,11 +177,19 @@ def run_download_scenario
       @template_output = sample_data('win_template_rendered_with_bootstrap_install_command.txt')
     end
 
-    it "bootstrap_install_command option is rendered in the windows-chef-client-msi.erb template" do
+    it "bootstrap_install_command option is rendered in the windows-chef-client-msi.erb template", :chef_lt_12_5_only => true do
```
btm: it'd be kind of nice if these were contexts too, like the 12.5.0+ below.

btm:
:+1:

PR #304, opened 2015-10-09
Feature: knife-windows should support --no-encryption flag

chefsalim:
This change add a new '--no-encryption' flag that can be specified with 'knife winrm' and 'knife bootstrap windows winrm'.  If this flag is set, it sets the transport to be 'plaintext', and turns off SSPI negotiation. This flag is not compatible with the winrm transport set to 'ssl' - it will emit an error and exit in that case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # knife-windows Change Log
 
+## Unreleased
+* New feature `--winrm-no-encryption` option for explicit control of WinRM encryption
+
 ## Release 1.0.0
 
 * [knife-windows #281](https://github.com/chef/knife-windows/pull/281) Prevent unencrypted negotiate auth, automatically prefix local usernames with '.' for negotiate
@@ -74,4 +77,3 @@ Release Notes - Knife Windows Plugin - Version 0.5.4
 
 ** New Feature
     * [KNIFE\_WINDOWS-6] - default bootstrap template should support encrypted\_data\_bag\_secret
-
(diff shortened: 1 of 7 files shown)


btm:
Is there an issue filed for this? What's the use case? Is this just a shortcut for `--winrm-authentication-protocol basic --winrm-transport plaintext` that also turns off any messages telling you that you shouldn't do that?

mwrock:
I think I have the same questions as @btm. It does not look like this alters functinoality and I think adding another argument to the mix makes the winrm auth story more confusing.

adamedx:
Here's the repro scenario -- let me know if it doesn't repro for you :) :+1: 

1. User has a Windows image where they have changed on default winrm setting -- the `wsman:localhost\service\allowunencrypted` to `true`
2. User has tries to use knife-windows to bootstrap it from a MacOS node using the plaintext protocol

If you try this with knife-windows 0.8.6, it works. But on 1.0.0, it won't, because knife-windows 1.0.0 raises an error that says using negotiate is not supported over plaintext on non-Windows systems.

adamedx:
Hmm, it looks like we don't actually fail anymore, we warn -- this was fixed by @mwrock. So the question is whether we should fail and require the specification of that flag, or just let the warning happen.

```WARNING: You are using '--winrm-authentication-protocol negotiate' with 
'--winrm-transport plaintext' on a non-Windows system which results in
unencrypted traffic. To avoid this warning and secure communication,
use '--winrm-transport ssl' instead of the plaintext transport,
or execute this command from a Windows system which enables encrypted
communication over plaintext with the negotiate authentication protocol.```

btm:
This turned out to effectively be a duplicate of #280, which has already been fixed.

PR #306, opened 2015-10-12
make knife-windows backwards compatible with most current chef clients

mwrock:
The policy file work introduced a new `validate_options!` method in `Knife::Bootstrap` that this calls in the `bootstrap` method. However, if using a pre 12.5 chef client, this method will not exist. So this PR adds that method to the bootstrap base module and simply calls `super` so if its in `Knife::Bootstrap` it will be called otherwise nothing happens.

### DIFF
--- a/lib/chef/knife/bootstrap_windows_base.rb
+++ b/lib/chef/knife/bootstrap_windows_base.rb
@@ -273,7 +273,9 @@ class Chef
         end
 
         validate_name_args!
-        validate_options!
+
+        # adding respond_to? so this works with pre 12.4 chef clients
+        validate_options! if respond_to?(:validate_options!)
 
         @node_name = Array(@name_args).first
         # back compat--templates may use this setting:


adamedx:
:+1:, I'm going to tell Travis to retry as the issue looks transient.

adamedx:
@danielsdeleo, @chefsalim, can you take a look?

danielsdeleo:
Not sure this will actually work, you might need to check `respond_to?` instead. Trying out code like this, I get a no method error:

```ruby
class A
  def do_super_to_object
    super
  end
end

class B < A
  def do_super_to_a
    super
  end
end

B.new.do_super_to_a
```

Gives me:

```
super.rb:9:in `do_super_to_a': super: no superclass method `do_super_to_a' for #<B:0x007fba6d08c760> (NoMethodError)
	from super.rb:13:in `<main>'
```

mwrock:
@danielsdeleo is definitely correct. I was initially testing with 12.4.3 and then switched to my chef in the gemfile. Somewhere along the way i installed 12.5.1 and so when i reverted the Gemfile I thought this worked. Switching to `respond_to`

mwrock:
@deepsihag @adamedx @chefsalim this is using `respond_to?1 now and I validated against chef `12.4.3`

danielsdeleo:
Might want to add a comment so someone in the future has an easier time knowing when they can remove the compatibility shim. :+1: otherwise

mwrock:
good point. added comment and squashed

PR #307, opened 2015-10-14
ensure prompted password is passed to winrm session

mwrock:
There is currently a bug in v1.0 that does not pass the prompted password to the winrm session.


adamedx:
@mwrock, curious about the cause -- was this introduced when we refactored a few things?

adamedx:
Glad to see we added a test to catch this.

mwrock:
yeah @adamedx  it got lost in the shuffle of option resolving refactoring

adamedx:
:+1:

ksubrama:
:+1:

PR #312, opened 2015-10-24
knife winrm bootstrap should not swallow failed exit codes

### LINKED ISSUE #311: WinRM bootstrap  silently fails
Story:
When trying to move from 0.8.5, to 1.1.0, we noticed the exit code for bootstrapping always succeeded, no matter if bootstrap fails or succeeds.  The knife winrm commands do not have this problem, since a different code path is used, this should also probably be part of the fix for this issue.  winrm/bootstrap should probably share the same basic relay_command code base.  We noticed with the [sample output](https://gist.github.com/carpnick/5486048ee6438734f833) we still got an exit code of 0

Diagnosis so far:
0.8.5
* When bootstrapping this [line](https://github.com/chef/knife-windows/blob/v0.8.5/lib/chef/knife/bootstrap_windows_winrm.rb#L63) gets executed when bootstrapping.
* This maps to [here](https://github.com/chef/knife-windows/blob/v0.8.5/lib/chef/knife/winrm.rb#L269).  If the command fails for any reason(except authentication) it [exits](https://github.com/chef/knife-windows/blob/v0.8.5/lib/chef/knife/winrm.rb#L293)
* Because it exits - it never gets back to [here](https://github.com/chef/knife-windows/blob/v0.8.5/lib/chef/knife/bootstrap_windows_base.rb#L181).  That means the code to return bootstrap_command_result was never being used.

1.1.0
* When bootstrapping this [line](https://github.com/chef/knife-windows/blob/v1.1.0/lib/chef/knife/bootstrap_windows_winrm.rb#L58) get executed.
* It runs then returns the exit code to [here](https://github.com/chef/knife-windows/blob/v1.1.0/lib/chef/knife/bootstrap_windows_base.rb#L323) and the code tries to return the exit code to knife.  Knife drops the return on the floor at this point.
* ASIDE - knife Winrm does not have this issue because it executes [this method](https://github.com/chef/knife-windows/blob/v1.1.0/lib/chef/knife/winrm.rb#L63) which has code near the bottom that [exits appropriately](https://github.com/chef/knife-windows/blob/v1.1.0/lib/chef/knife/winrm.rb#L71-L79).  This is a major problem because the 2 code bases dont share the code to determine if the command failed.  But it does share the common WinRMSession code.  Suggest this gets fixed.

knife bootstrap (ssh) (12.5.1)
* Runs bootstrap via ssh [here](https://github.com/chef/chef/blob/12.0.3/lib/chef/knife/bootstrap.rb#L256)
* Looking at the SSH [method](https://github.com/chef/chef/blob/12.0.3/lib/chef/knife/ssh.rb#L485) it [exits](https://github.com/chef/chef/blob/12.0.3/lib/chef/knife/ssh.rb#L518) appropriately.

So it looks like the change that is needed is to change [this line](https://github.com/chef/knife-windows/blob/v1.1.0/lib/chef/knife/bootstrap_windows_base.rb#L325) to exit the bootstrap_command_result.  That solves the immediate problem.  However it should be noted that the code to execute commands should probably be common within knife-windows to determine success/failure.

mwrock:
This fixes #311 and consolidates error handling logic to ensure both knife winrm and knife winrm bootstrap follow the same path.

### DIFF
--- a/lib/chef/knife/winrm_knife_base.rb
+++ b/lib/chef/knife/winrm_knife_base.rb
@@ -25,6 +25,14 @@ require 'chef/knife/knife_windows_base'
 class Chef
   class Knife
     module WinrmCommandSharedFunctions
+
+      FAILED_BASIC_HINT ||= "Hint: Please check winrm configuration 'winrm get winrm/config/service' AllowUnencrypted flag on remote server."
+      FAILED_NOT_BASIC_HINT ||= <<-eos.gsub /^\s+/, ""
+        Hint: Make sure to prefix domain usernames with the correct domain name.
+        Hint: Local user names should be prefixed with computer name or IP address.
+        EXAMPLE: my_domain\\user_namer
+      eos
+
       def self.included(includer)
         includer.class_eval do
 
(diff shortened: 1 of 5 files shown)


mwrock:
Validated a failed bootstrap has the correct `$LASTEXITCODE`

btm:
:+1:

lib/chef/knife/winrm_knife_base.rb
```diff
@@ -93,8 +101,91 @@ def resolve_target_nodes
...
+          def run_command(command = '')
+            relay_winrm_command(command)
+
+            check_for_errors!
+
+            # Knife seems to ignore the return value of this method,
+            # so we exit to force the process exit code for this
+            # subcommand if returns is set
+            exit @exit_code if @exit_code && @exit_code != 0
+            @exit_code || 0
```
jaym: should this just be 0?

jaym:
:+1:

lib/chef/knife/winrm_knife_base.rb
mwrock: yup

PR #329, opened 2015-12-30
pin to a minimum winrm-s of 0.3.2

mwrock:


jaym:
:+1:

knife-windows.gemspec
```diff
@@ -15,7 +15,7 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version	= ">= 1.9.1"
   s.add_dependency "winrm", "~> 1.3"
-  s.add_dependency "winrm-s", "~> 0.3.0"
+  s.add_dependency "winrm-s", "~> 0.3.0", ">= 0.3.2"
```
btm: ` s.add_dependency "winrm-s", "~> 0.3.2"` ?

btm:
:+1: to get people https://github.com/chef/winrm-s/pull/29, but I think you can get by with one constraint there.

knife-windows.gemspec
mwrock: right. i was thinking of the ~> 0.3 being the first constraint which clearly it is not. will change to what you have above

PR #334, opened 2016-01-21
drop winrm-s dependency and allow both windows and linux to use negotiate auth

mwrock:
This is a WIP PR dependent on the following downstream PRs:
https://github.com/WinRb/rubyntlm/pull/26
https://github.com/WinRb/WinRM/pull/144

The above PRs allow the winrm gem to provide sspinegotiate authentication and eliminates the need for the winrm-s monkey patches. the winrm changes work with the current httpclent API as is injecting the appropriate headers so no patches needed for httpclient. Further, the winrm changes leverage the rubyntlm gem which implements negotiate authentication in pure ruby. This allows us to leverage negotiate on linux workstations and do away with the OS checks and warnings regarding plaintext on linux.

cc @sneal @zenchild

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,9 @@ source "https://rubygems.org"
 # Specify your gem's dependencies in knife-windows.gemspec
 gemspec
 
+gem 'winrm', git: 'https://github.com/WinRb/WinRM', branch: 'dan/ntlm-encryption-squashed'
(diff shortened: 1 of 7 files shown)


randomcamel:
I'm not a domain expert, but LGTM. :+1:

adamedx:
:+1:. We've removed @btm's helpful warning since negotiate now works universally, and I think that's appropriate.

mwrock:
Just to note, this has been tested against:
- 2012R2
- 2008 R2
- Nano

PR #340, opened 2016-01-26
Ignore the FIPS flag in tests

jaym:
The fips flag is causing tests to fail because it is expecting it
in 12.6, however it's only in current master https://github.com/chef/chef/commit/58112e1210cac991e7a9cf434c74ece0aa97414d
and will be part of the 12.7 release.


mwrock:
:+1:

PR #341, opened 2016-01-27
fixes wsman test and death to nokogiri

mwrock:
This makes the following fixes to `wsman test`:
* Do not swallow deeper http errors but send to verbode
* Do not prompt for password
* honor `--winrm-ssl-verify-mode verify_none`
* replace nokogiri with rexml


danielsdeleo:
:+1:

randomcamel:
:+1:

PR #345, opened 2016-01-28
add more guidance around possible ssl verification issues

mwrock:
This change attempts to take some more "guess work" out of SSL verification errors. It also refactors the huge `verify_wsman_accessiblity_for_nodes` to smaller methods.

### DIFF
--- a/lib/chef/knife/wsman_test.rb
+++ b/lib/chef/knife/wsman_test.rb
@@ -54,10 +54,13 @@ class Chef
           }
           output_object = Chef::Knife::WsmanEndpoint.new(item.host, item.port, item.endpoint)
           error_message = nil
+          ssl_error = nil
           begin
             client = HTTPClient.new
             client.ssl_config.verify_mode = OpenSSL::SSL::VERIFY_NONE if resolve_no_ssl_peer_verification
             response = client.post(item.endpoint, xml, header)
+          rescue OpenSSL::SSL::SSLError => ssl_error
+            error_message = ssl_error.message
           rescue Exception => e
             error_message = e.message
           else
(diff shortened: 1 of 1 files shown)


lib/chef/knife/wsman_test.rb
```diff
@@ -42,44 +42,29 @@ def run
...
-          end
+          output_object = parse_response(item, response)
+          output_object.error_message += "\r\nError returned from endpoint: #{err.message}" if err
 
-          unless error_message.nil?
+          unless output_object.error_message.nil?
             ui.warn "Failed to connect to #{item.host} at #{item.endpoint}."
-            output_object.error_message = error_message
+            if err && err.is_a?(OpenSSL::SSL::SSLError)
+              ui.warn "Failure due to an issue with SSL likely cause would be unsuccesful certificate verification."
```
randomcamel: Should be a semicolon after "SSL". (Small thing, but a nice detail.)

randomcamel:
:+1: with additional punctuation.

lib/chef/knife/wsman_test.rb
```diff
@@ -42,44 +42,29 @@ def run
...
+          output_object = parse_response(item, response)
+          output_object.error_message += "\r\nError returned from endpoint: #{err.message}" if err
 
-          unless error_message.nil?
+          unless output_object.error_message.nil?
             ui.warn "Failed to connect to #{item.host} at #{item.endpoint}."
-            output_object.error_message = error_message
+            if err && err.is_a?(OpenSSL::SSL::SSLError)
+              ui.warn "Failure due to an issue with SSL; likely cause would be unsuccesful certificate verification."
+              ui.warn "Either ensure your certificate is valid or use '--winrm-ssl-verify-mode verify_none' to ignore verification failures."
```
danielsdeleo: Can `knife ssl check` run against a winrm service? I think it should be able to...

danielsdeleo: Recapping from video chat about this: `knife ssl check` does some nifty things that we might want to emulate, but the messages emitted by that command all refer very specifically to Chef Server (for example, telling you where to get the default self-signed cert from your chef server's file system) so having the user run `knife ssl check` against a winrm endpoint would be confusing to users. It would be good if we could make this command tell the difference between lack of CA trust vs. hostname mismatch, but we can do that in the future.

danielsdeleo:
:+1:

PR #347, opened 2016-02-04
Knife bootstrap includes `fips true` when in fips mode

jaym:
If knife has set fips mode, make sure to add it to
the client.rb during bootstraps

Tested with basic auth + ssl transport

Some notes:
* ntlm authentication does not work. It uses MD4 and RC4
  at some point. When asking openssl to do any of these
  operations at any point while in fips mode, ruby will
  refuse the request and raise. What this means is that
  users will have to use basic auth over ssl or potentially
  kerberos.


mwrock:
:+1: on green

ksubrama:
:+1: Go for it.

PR #352, opened 2016-02-25
Have client.rb verify that FIPS mode can be enforced

ksubrama:


ksubrama:
@jaym @btm @chef/client-windows 

This mirrors https://github.com/chef/chef/pull/4630

mwrock:
:+1:

btm:
:+1:

hh:
I wasn't sure what FIPS was, so a bit of googling brought up:

http://blogs.technet.com/b/secguide/archive/2014/04/07/why-we-re-not-recommending-fips-mode-anymore.aspx

Looks like it's not recommended anymore, nice to have for completeness.

ksubrama:
@hh I do not believe you would ever accidentally encounter FIPS without it being a compliance requirement.  FIPS-140 is a NIST standard for infosec/crypto when handling government related data and comes in various flavors and with varyingly hilarious implications.  It's an absolute necessity for those in that field, regardless of its technical merit, and very much not a thing for those outside it.

This change only affects a tiny number of platforms and only when explicitly requested.  It will never be the default for regular usage - so you can mostly ignore it.

mwrock:
I'd like to ask that we all bow our heads in a moment of silence thinking of those who have this requirement.

PR #354, opened 2016-03-04
Adding the ability for users to specify the architecture (64 or 32 bit) of Chef that they want installed during bootstrapping

tyler-ball:
This is ready for review.

\cc @smurawski @mwrock @adamedx @chefsalim @randomcamel

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,9 +58,9 @@ SSL will become the default transport in future revisions of
 
 #### Specifying the package architecture
 
-You can configure which package architecture (32 bit or 64 bit) to install on the bootstrapped system.  In your knife config specify `knife[:bootstrap_architecture]`.  Valid values are `:i386` for 32 bit or `x86_64` for 64 bit.  By default the architecture will be whatever the target system is.  If you try to install a 64 bit package on a 32 bit system you will receive an error, but installing a 32 bit package on a 64 bit system is supported.
+You can configure which package architecture (32 bit or 64 bit) to install on the bootstrapped system.  In your knife config specify `knife[:bootstrap_architecture]`.  Valid values are `:i386` for 32 bit or `:x86_64` for 64 bit.  By default the architecture will be whatever the target system is.  If you try to install a 64 bit package on a 32 bit system you will receive an error, but installing a 32 bit package on a 64 bit system is supported.
 
-Currently (March 2016) the `stable` channel of omnibus (where downloads using the install script fetch) only has 32 bit packages but this will be updated soon to include both 32 and 64 bit packages.  Until then you will need to access the `current` channel by specifying `--prerelease` in your `knife bootstrap windows`.
+Currently (March 2016) the `stable` channel of omnibus (where downloads using the install script fetch) only has 32 bit packages but this will be updated soon to include both 32 and 64 bit packages.  Until then you will need to access the `current` channel by specifying `--prerelease` in your `knife bootstrap windows` if you want 64 bit packages.
 
 ### knife wsman test
 
--- a/README.md
+++ b/README.md
@@ -56,6 +56,12 @@ support](#platform-winrm-authentication-support).
 SSL will become the default transport in future revisions of
 `knife-windows`.
 
+#### Specifying the package architecture
+
+You can configure which package architecture (32 bit or 64 bit) to install on the bootstrapped system.  In your knife config specify `knife[:bootstrap_architecture]`.  Valid values are `:i386` for 32 bit or `x86_64` for 64 bit.  By default the architecture will be whatever the target system is.  If you try to install a 64 bit package on a 32 bit system you will receive an error, but installing a 32 bit package on a 64 bit system is supported.
+
+Currently (March 2016) the `stable` channel of omnibus (where downloads using the install script fetch) only has 32 bit packages but this will be updated soon to include both 32 and 64 bit packages.  Until then you will need to access the `current` channel by specifying `--prerelease` in your `knife bootstrap windows`.
+
 ### knife wsman test
 
 Connects to the remote WSMan/WinRM endpoint and verifies the remote node is listening.  This is the equivalent of running Test-Wsman from PowerShell.  Endpoints to test can be specified manually, or be driven by search and use many of the same connection options as knife winrm.
@@ -125,7 +131,7 @@ which does the following:
 * Writes a default config file for Chef (`C:\chef\client.rb`) using values from the `knife.rb`.
 * Creates a JSON attributes file containing the specified run list and run Chef.
 
-This template is used by both `knife windows bootstrap winrm` and `knife windows bootstrap ssh` subcommands.
+This template is used by both `knife bootstrap windows winrm` and `knife bootstrap windows ssh` subcommands.
 
 ## Requirements / setup
 
--- a/lib/chef/knife/bootstrap_windows_base.rb
+++ b/lib/chef/knife/bootstrap_windows_base.rb
@@ -278,6 +278,11 @@ class Chef
           warn_chef_config_secret_key
         end
 
+        bootstrap_architecture = Chef::Config[:knife][:bootstrap_architecture]
+        if bootstrap_architecture && ![:x86_64, :i386].include?(bootstrap_architecture.to_sym)
+          raise "Valid values for the knife config :bootstrap_architecture are i386 or x86_64.  Supplied value is #{bootstrap_architecture}"
+        end
+
         validate_name_args!
 
         # adding respond_to? so this works with pre 12.4 chef clients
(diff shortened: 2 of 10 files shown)


lib/chef/knife/bootstrap_windows_base.rb
```diff
@@ -319,6 +322,23 @@ def bootstrap(proto=nil)
         end
 
         wait_for_remote_response( config[:auth_timeout].to_i )
+
+        # We allow the user to specify the desired architecture of Chef to install or we default
+        # to whatever the target system is
+        session_results = relay_winrm_command("echo %PROCESSOR_ARCHITECTURE%")
+        current_architecture = session_results[0][:data][0][:stdout] == "X86" ? :i386 : :x86_64
```
mwrock: We should examine each potential node separately instead of basing the architecture for all nodes from the results of the first.

tyler-ball: But is it possible to bootstrap more than 1 node at once?  I didn't think so

mwrock: Oh duh...I was thinking khife winrm. Never mind.

lib/chef/knife/bootstrap_windows_base.rb
```diff
@@ -319,6 +323,28 @@ def bootstrap(proto=nil)
         end
 
         wait_for_remote_response( config[:auth_timeout].to_i )
+
+        # We allow the user to specify the desired architecture of Chef to install or we default
+        # to whatever the target system is.  We assume that we are only bootstrapping 1 node at a time
+        # so we don't need to worry about multipe responses from this command.
+        session_results = relay_winrm_command("echo %PROCESSOR_ARCHITECTURE%")
+        # TODO is this valid error checking?
```
tyler-ball: @mwrock What do you think about the error checking I'm doing here?  I'm not sure what kind of responses I will get from `relay_winrm_command` so If its at all different from what I expect I raise an error.

mwrock: yeah you can clean it up a bit. Each `session_results` will be a [`WinRM::Output`](https://github.com/WinRb/WinRM/blob/master/lib/winrm/output.rb) instance with an `stdout` method. So:
```
if session_results[0].data.empty?
```
should suffice.

README.md
```diff
@@ -56,6 +56,12 @@ support](#platform-winrm-authentication-support).
 SSL will become the default transport in future revisions of
 `knife-windows`.
 
+#### Specifying the package architecture
+
+You can configure which package architecture (32 bit or 64 bit) to install on the bootstrapped system.  In your knife config specify `knife[:bootstrap_architecture]`.  Valid values are `:i386` for 32 bit or `x86_64` for 64 bit.  By default the architecture will be whatever the target system is.  If you try to install a 64 bit package on a 32 bit system you will receive an error, but installing a 32 bit package on a 64 bit system is supported.
+
+Currently (March 2016) the `stable` channel of omnibus (where downloads using the install script fetch) only has 32 bit packages but this will be updated soon to include both 32 and 64 bit packages.  Until then you will need to access the `current` channel by specifying `--prerelease` in your `knife bootstrap windows`.
```
btm: add "if you want 64bit packages" to the end of this?

README.md
```diff
@@ -56,6 +56,12 @@ support](#platform-winrm-authentication-support).
 SSL will become the default transport in future revisions of
 `knife-windows`.
 
+#### Specifying the package architecture
+
+You can configure which package architecture (32 bit or 64 bit) to install on the bootstrapped system.  In your knife config specify `knife[:bootstrap_architecture]`.  Valid values are `:i386` for 32 bit or `x86_64` for 64 bit.  By default the architecture will be whatever the target system is.  If you try to install a 64 bit package on a 32 bit system you will receive an error, but installing a 32 bit package on a 64 bit system is supported.
```
btm: One of the values is presented as a symbol, the other isn't.

lib/chef/knife/bootstrap_windows_base.rb
```diff
@@ -319,6 +324,26 @@ def bootstrap(proto=nil)
...
+        if bootstrap_architecture.nil?
+          architecture = current_architecture
+        elsif bootstrap_architecture == :x86_64 && current_architecture == :i386
+          raise "You specified bootstrap_architecture as x86_64 but the target machine is i386.  A 64 bit program cannot run on a 32 bit machine."
+        else
+          architecture = bootstrap_architecture
+        end
+        # The windows install script wants i686, not i386
+        architecture = :i686 if architecture == :i386
+        Chef::Config[:knife][:architecture] = architecture
```
btm: Check to see if `Chef::Config[:knife][:architecture]` is already set somewhere and raise telling the user to set `Chef::Config[:knife][:bootstrap_architecture]` instead?

lib/chef/knife/bootstrap_windows_base.rb
```diff
@@ -319,6 +324,26 @@ def bootstrap(proto=nil)
         end
 
         wait_for_remote_response( config[:auth_timeout].to_i )
+
```
mwrock: While the bootstrap method is already too long. It would be a good ides to extract this bit to a separate method.

lib/chef/knife/bootstrap_windows_base.rb
btm: I have a hard time keeping track of the difference between current_architecture, bootstrap_architecture, and architecture here. After I read it a few times, I stopped getting confused. So I don't know. naming is hard.

spec/unit/knife/bootstrap_windows_winrm_spec.rb
```diff
@@ -262,6 +284,7 @@
     let(:node_name) { 'foo.example.com' }
     before do
       allow(bootstrap).to receive(:wait_for_remote_response)
+      expect(bootstrap).to receive(:relay_winrm_command).with("echo %PROCESSOR_ARCHITECTURE%").and_return(arch_session_results)
```
btm: this isn't being tested, so shouldn't this be an allow?

spec/unit/knife/bootstrap_windows_winrm_spec.rb
```diff
@@ -187,6 +208,7 @@
   end
 
   it 'should have a wait timeout of 2 minutes by default' do
+    expect(bootstrap).to receive(:relay_winrm_command).with("echo %PROCESSOR_ARCHITECTURE%").and_return(arch_session_results)
```
btm: this isn't being tested, so shouldn't this be an allow?

btm:
test for the i686/i386 switcheroo?

README.md
tyler-ball: :+1:

README.md
tyler-ball: :+1:

lib/chef/knife/bootstrap_windows_base.rb
tyler-ball: :+1:

lib/chef/knife/bootstrap_windows_base.rb
tyler-ball: extracted

lib/chef/knife/bootstrap_windows_base.rb
```diff
@@ -319,6 +324,26 @@ def bootstrap(proto=nil)
...
+        # We allow the user to specify the desired architecture of Chef to install or we default
+        # to whatever the target system is.  We assume that we are only bootstrapping 1 node at a time
+        # so we don't need to worry about multipe responses from this command.
+        session_results = relay_winrm_command("echo %PROCESSOR_ARCHITECTURE%")
+        if session_results.empty? || session_results[0].stdout.strip.empty?
+          raise "Response to 'echo %PROCESSOR_ARCHITECTURE%' command was invalid: #{session_results}"
+        end
+        current_architecture = session_results[0].stdout.strip == "X86" ? :i386 : :x86_64
```
chefsalim: whitespace

lib/chef/knife/bootstrap_windows_base.rb
```diff
@@ -319,6 +324,26 @@ def bootstrap(proto=nil)
...
+        if bootstrap_architecture.nil?
+          architecture = current_architecture
+        elsif bootstrap_architecture == :x86_64 && current_architecture == :i386
+          raise "You specified bootstrap_architecture as x86_64 but the target machine is i386.  A 64 bit program cannot run on a 32 bit machine."
+        else
+          architecture = bootstrap_architecture
+        end
+        # The windows install script wants i686, not i386
```
chefsalim: whitespace

spec/unit/knife/bootstrap_windows_winrm_spec.rb
tyler-ball: I always prefer expect to allow because it prevents bloat from forming in the specs over time as code is changed.  But if you would prefer, I will change it?

lib/chef/knife/bootstrap_windows_base.rb
chefsalim: would it be more confusing or less to just standardize on ':i686' instead of ':i386' for bootstrap_architecture value?

tyler-ball:
@btm added a test for the switcheroo

tyler-ball:
@mwrock @btm review comments addressed in the latest commit

lib/chef/knife/bootstrap_windows_base.rb
tyler-ball: Addressed

lib/chef/knife/bootstrap_windows_base.rb
```diff
@@ -410,6 +421,29 @@ def warn_chef_config_secret_key
...
+        session_results = relay_winrm_command("echo %PROCESSOR_ARCHITECTURE%")
+        if session_results.empty? || session_results[0].stdout.strip.empty?
+          raise "Response to 'echo %PROCESSOR_ARCHITECTURE%' command was invalid: #{session_results}"
+        end
+        current_architecture = session_results[0].stdout.strip == "X86" ? :i386 : :x86_64
+
+        if bootstrap_architecture.nil?
+          architecture = current_architecture
+        elsif bootstrap_architecture == :x86_64 && current_architecture == :i386
+          raise "You specified bootstrap_architecture as x86_64 but the target machine is i386.  A 64 bit program cannot run on a 32 bit machine."
```
chefsalim: Remove extra space after the '.'

lib/chef/knife/bootstrap_windows_base.rb
```diff
@@ -410,6 +421,29 @@ def warn_chef_config_secret_key
...
+        else
+          architecture = bootstrap_architecture
+        end
+
+        # The windows install script wants i686, not i386
```
chefsalim: Alternatively, it might be less confusing to just do this conversion in the .bat (.erb) file

mwrock:
I approve this pr :+1:

chefsalim:
The changes seem ok and I'm fine with taking them.  However, I do want to point out the strategy of making an extra winrm call, and the associated changes seem like they add more complexity than is needed.  It could be simpler to just add the desired architecture to the erb file, and let it do the checks and proceed appropriately.

PR #366, opened 2016-04-30
Update to same rubies that chef uses

mwrock:
travis and appveyor are failing due to chef-zero requiring ruby 2.1.0. Since chef only tests against 2.1 and 2.2 now, we'll do the same here.

I'm also pinning bundler since the latest is broken on windows but a new release is imminent: https://github.com/bundler/bundler/pull/4479#issuecomment-215901069

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: ruby
 
 rvm:
-  - 1.9.3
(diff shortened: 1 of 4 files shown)


randomcamel:
👍

PR #362, opened 2016-04-26
Adding a proxy env var to wget.ps1

tristanmorgan:
Hi, I'm trying to get a machine to bootstrap from behind a proxy with knife[:bootstrap_proxy] set in my knife.rb and the powershell script didn't pick up the environment vars.

PR #367, opened 2016-05-03
Honor the chef ssl_policy when making winrm calls

mwrock:
Currently it is not possible for winrm calls to validate ssl certificates in the user's `trusted_certificates` folder. This has confused several users because while calls to `knife ssl check` leads ione to believe that a ssl certificate is trusted, subsequent bootstraps and `knife winrm` calls fail unless they turn off ssl validation.

This applies the `Chef::HTTP::DefaultSSLPolicy` to the winrm `HTTPClient` connection and therefore honors all certificates in the trusted_certs folder.

Admittedly the implementation here is quite hacky as it has to get to the `HTTPClient` via direct instance variable inspection. While I could expose those in winrm, that would require a minor version release and I'm trying to keep winrm releases at bay in order to focus on winrm v2.

### DIFF
--- a/spec/dummy_winrm_service.rb
+++ b/spec/dummy_winrm_service.rb
@@ -1,0 +1,24 @@
+module Dummy
+  class WinRMTransport
+    attr_reader :httpcli
+
+    def initialize
+      @httpcli = HTTPClient.new
+    end
+  end
+
+  class WinRMService
+    attr_reader :xfer
+
+    def initialize
+      @xfer = WinRMTransport.new
+    end
+
+    def set_timeout(timeout); end
+    def open_shell; end
+    def run_command; end
+    def get_command_output; end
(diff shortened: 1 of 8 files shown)


randomcamel:
👍

spec/dummy_winrm_service.rb
```diff
@@ -0,0 +1,24 @@
+module Dummy
```
ksubrama: Are you doing this because of some lulz when combining instance doubles with instance_variable_get?

mwrock: yeah because the `HTTPClient` is a couple classes deep, accessing via a double got awkward and this makes it easier to share in different tests.

mwrock:
waiting for one more +1 on #362 and then I'll edit the changelog and release.

PR #368, opened 2016-05-03
fix tests with fips check

mwrock:
This is needed for extenal tests in https://github.com/chef/chef/pull/4880 to pass


randomcamel:
👍

PR #377, opened 2016-07-29
Added code and corresponding RSpecs to read the json attributes from the --json-attributes-file option.

Aliasgar16:
Done. json attributes passed through --json-attributes-file option were not getting set into config object and so the attributes passed were then missing from the first_boot.json file.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,4 +9,5 @@ group :test do
   gem "ruby-wmi"
   gem "httpclient"
   gem 'rake'
+  gem "rack", "< 2.0" # 2.0 requires Ruby 2.2+
 end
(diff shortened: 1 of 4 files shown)


Aliasgar16:
Pinned `rack gem` version in `ci.gemfile` as well.

mwrock:
:+1:

btm:
Can you use `let` instead of instance variables in the tests?

http://stackoverflow.com/questions/5359558/when-to-use-rspec-let

e.g.
```
let(:first_boot_attributes) { { 'a1' => 'b1', 'a2' => 'b2' } }
```

Aliasgar16:
Fixed the review comments.

Issue #372, opened 2016-06-08
knife bootstrap through ssh fails


anuriq:
I am using FreeSSHd on windows nodes. Everything worked fine until newest release of Chef DK.

Error:
/opt/chefdk/embedded/lib/ruby/gems/2.1.0/gems/knife-windows-1.4.1/lib/chef/knife/bootstrap_windows_base.rb:429:in `set_target_architecture': undefined method `relay_winrm_command' for #<Chef::Knife::BootstrapWindowsSsh:0x007fb8cc319de0> (NoMethodError)

Full debug log available [here](https://gist.github.com/anuriq/04bf5d6d5eb3749a68e83c41a5edfd68)

anuriq:
btw bootstrap through winrm works fine.

peterdeme:
Hi anuriq

I faced the same issue. Not sure what relay_winrm_command do but instead I manually modified the os_archtiectures like this:

```
 def set_target_architecture(bootstrap_architecture)
     
        current_architecture = : :x86_64

        if bootstrap_architecture.nil?
          architecture = current_architecture
        elsif bootstrap_architecture == :x86_64 && current_architecture == :i386
          raise "You specified bootstrap_architecture as x86_64 but the target machine is i386. A 64 bit program cannot run on a 32 bit machine."
        else
          architecture = bootstrap_architecture
        end

        # The windows install script wants i686, not i386
        architecture = :i686 if architecture == :i386
        Chef::Config[:knife][:architecture] = architecture
```

you have to use sudo to modify that ruby file.

mwrock:
ugh. this is definitely a new bug that breaks ssh based bootstrap on windows.

anuriq:
A call to `relay_winrm_command` should not be in `bootstrap_windows_base.rb`. Maybe it was meant to be in `bootstrap_windows_winrm.rb`. I'd sent a PR if I knew what was the intention...

mwrock:
I think it was just added here not thinking of ssh. The fact is that few use this gem using ssh and so it often is forgotten about. We should really have some integration tests to protect against this. Anyways yeah I think the right approach would be to abstract away the transport and have `bootstrap_windows_base.rb` call that abstraction instead of a specific transport.

mwrock:
fixed via #380

PR #380, opened 2016-08-09
Fixing bootstrap via ssh regression

NimishaS:
w.r.t. https://github.com/chef/knife-windows/issues/372

### DIFF
--- a/lib/chef/knife/bootstrap_windows_base.rb
+++ b/lib/chef/knife/bootstrap_windows_base.rb
@@ -431,11 +431,15 @@ behavior will be removed and any 'encrypted_data_bag_secret' entries in
       # to whatever the target system is.  We assume that we are only bootstrapping 1 node at a time
       # so we don't need to worry about multipe responses from this command.
       def set_target_architecture(bootstrap_architecture)
-        session_results = relay_winrm_command("echo %PROCESSOR_ARCHITECTURE%")
-        if session_results.empty? || session_results[0].stdout.strip.empty?
-          raise "Response to 'echo %PROCESSOR_ARCHITECTURE%' command was invalid: #{session_results}"
+        if (self.class == Chef::Knife::BootstrapWindowsWinrm)
+          session_results = relay_winrm_command("echo %PROCESSOR_ARCHITECTURE%")
+          if session_results.empty? || session_results[0].stdout.strip.empty?
+            raise "Response to 'echo %PROCESSOR_ARCHITECTURE%' command was invalid: #{session_results}"
+          end
+          current_architecture = session_results[0].stdout.strip == "X86" ? :i386 : :x86_64
+        else
+          current_architecture =  :x86_64
         end
-        current_architecture = session_results[0].stdout.strip == "X86" ? :i386 : :x86_64
 
         if bootstrap_architecture.nil?
           architecture = current_architecture
--- a/lib/chef/knife/bootstrap_windows_base.rb
+++ b/lib/chef/knife/bootstrap_windows_base.rb
@@ -283,6 +283,8 @@ class Chef
           warn_chef_config_secret_key
         end
 
+        set_target_architecture
+
         validate_name_args!
 
         # adding respond_to? so this works with pre 12.4 chef clients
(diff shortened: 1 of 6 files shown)


lib/chef/knife/bootstrap_windows_base.rb
```diff
@@ -431,11 +431,15 @@ def warn_chef_config_secret_key
...
-        if session_results.empty? || session_results[0].stdout.strip.empty?
-          raise "Response to 'echo %PROCESSOR_ARCHITECTURE%' command was invalid: #{session_results}"
+        if (self.class == Chef::Knife::BootstrapWindowsWinrm)
+          session_results = relay_winrm_command("echo %PROCESSOR_ARCHITECTURE%")
+          if session_results.empty? || session_results[0].stdout.strip.empty?
+            raise "Response to 'echo %PROCESSOR_ARCHITECTURE%' command was invalid: #{session_results}"
+          end
+          current_architecture = session_results[0].stdout.strip == "X86" ? :i386 : :x86_64
+        else
+          current_architecture =  :x86_64
```
NimishaS: This can't be hard-coded and will be modified.

mwrock: Instead of "pulling" the architecture from the node, consider moving this directly into the generated .bat file.

lib/chef/knife/bootstrap_windows_base.rb
```diff
@@ -283,14 +283,6 @@ def bootstrap(proto=nil)
...
 
-        bootstrap_architecture = Chef::Config[:knife][:bootstrap_architecture]
-        if bootstrap_architecture && ![:x86_64, :i386].include?(bootstrap_architecture.to_sym)
-          raise "Valid values for the knife config :bootstrap_architecture are i386 or x86_64. Supplied value is #{bootstrap_architecture}"
-        end
-        if Chef::Config[:knife][:architecture]
-          raise "Do not set :architecture in your knife config, use :bootstrap_architecture."
-        end
-
         validate_name_args!
```
NimishaS: Not sure why the user should give architecture as an input when every server will have a fixed architecture and we can find it ourselves. Taking input from the user increases the risk of passing wrong value too.
Hence removed this.

lib/chef/knife/bootstrap_windows_base.rb
```diff
@@ -426,29 +416,6 @@ def warn_chef_config_secret_key
...
-        if bootstrap_architecture.nil?
-          architecture = current_architecture
-        elsif bootstrap_architecture == :x86_64 && current_architecture == :i386
-          raise "You specified bootstrap_architecture as x86_64 but the target machine is i386. A 64 bit program cannot run on a 32 bit machine."
-        else
-          architecture = bootstrap_architecture
-        end
-
-        # The windows install script wants i686, not i386
-        architecture = :i686 if architecture == :i386
```
NimishaS: https://docs.chef.io/api_omnitruck.html says that the valid value of `m (architecture)` is `i386` for windows. But since the comment here says that it should be `i686`, using the same in `windows-chef-client-msi.erb`

lib/chef/knife/bootstrap_windows_base.rb
mwrock: They may want to install 32 bit chef on a 64bit architecture.

btm: :+1: for what @mwrock said. We used to only provide 32-bit builds, which still work on all 64-bit Windows servers except Nano. There are some complications that would make someone stick with 32-bit until they were ready to deal with them, like file / registry redirection.

https://blog.chef.io/2015/11/19/pre-release-announcement-native-64-bit-windows-chef-client/

lib/chef/knife/bootstrap_windows_base.rb
btm: I think we've standardized on i386 in omnitruck, but there's some back-compat support for i686 too. it looks like #msi_url in knife-windows hits the omnitrucks API rather than constructing a URL (which is correct) so it shouldn't matter. 

https://github.com/chef/omnitruck/blob/master/app.rb#L314

But maybe there's some old functionality I'm missing. we should test a few versions either way, like something before 12.9. Before 12.9 IIRC we always server 32-bit packages out of stable, but you could get 64bit packages out of current if they existed. For 12.9+ you should get whatever you ask for.

NimishaS:
Testing is in progress. @btm , @mwrock , shared the spreadsheet of testing.

Need to fix the specs too.

NimishaS:
@btm , @mwrock I have updated the testing sheet with some of the scenarios that are failing. Most of them are not regressions. Please take a look.

mwrock:
Just added comments to the spreadsheet. I think I think this was reported in a recent knife-windows issue due to casing. See https://github.com/chef/knife-windows/issues/382 informs 2 of the issues.

NimishaS:
@mwrock , @btm , please review the changes.  The failing cases reported in the sheet are not regressions. We can fix them as separate issues. 
Some specs have been removed in this PR as the corresponding ruby code has been removed while moving the logic to find machine architecture from ruby to .bat file

mwrock:
LGTM :+1:

btm:
👍

Issue #382, opened 2016-08-12
Windows arch detection during bootstrap is incorrect for 32-bit


glasschef:
Windows arch detection during bootstrap checks for `X86` when it's actually `x86` (in 32-bit Windows 7, at least, possibly other versions) so it falls back to 64-bit install, which can't install:

https://github.com/chef/knife-windows/blob/710bceb8c7c17cdde8f6f59591add920ce1d9d33/lib/chef/knife/bootstrap_windows_base.rb#L438

## knife-windows Version

1.5.0

## Platform Version

Windows 7 32-bit

## Replication Case

Attempt a `knife windows bootstrap winrm` of 32bit Windows with the knife-windows plugin 1.5.0

## Client Output

```
DEBUG: 10.211.192.138[552B1929-8A47-4272-A6D8-A933EA4BC9EF] => :run_command[echo %PROCESSOR_ARCHITECTURE%]
10.211.192.138 x86

...

10.211.192.138 powershell.exe -ExecutionPolicy Unrestricted -NoProfile -NonInteractive -File  C:\chef\wget.ps1 "https://www.chef.io/chef/download?p=windows&pv=2008r2&m=x86_64&DownloadContext=PowerShell&v=12" "C:\Users\clhadmin\AppData\Local\Temp\chef-client-latest.msi"

```
## Stacktrace

```
10.211.192.155 C:\Users\clhadmin>msiexec /qn /log "C:\Users\clhadmin\AppData\Local\Temp\chef-client-msi5504.log" /i "C:\Users\clhadmin\AppData\Local\Temp\chef-client-latest.msi"
10.211.192.155 T h i s   i n s t a l l a t i o n   p a c k a g e   i s   n o t   s u p p o r t e d   b y   t h i s   p r o c e s s o r   t y p e .   C o n t a c t   y o u r   p r o d u c t   v e n d o r .
10.211.192.155
```

NimishaS:
This issue gets fixed in https://github.com/chef/knife-windows/pull/380

charlesjohnson:
Fixed via #380, #384.

PR #384, opened 2016-08-18
Fix for architechture detection issue for 64 bit

dheerajd-msys:
During knife windows bootstrapping of windows 64 bit machine using `ssh` protocol, the code is not able to determine the  architecture properly, `PROCESSOR_ARCHITECTURE` value returns `x86(32bit)` instead of `AMD64(64 bit)`.
The behaviour is due to WOW64, added more validation to detect the architecture properly.

Following are values for various scenarios:- 

        Environment Variable      	 | 32bit Native	| 64bit Native	|  WOW64
        PROCESSOR_ARCHITECTURE   	 |   x86        |    AMD64      |  x86
        PROCESSOR_ARCHITEW6432     	 | undefined    |   undefined	|  AMD64

Reference:- https://blogs.msdn.microsoft.com/joshwil/2004/03/11/as-promised-what-is-the-wow64-and-what-does-it-mean-to-managed-apps-that-you-run-on-64bit-machines/

### DIFF
--- a/lib/chef/knife/bootstrap/windows-chef-client-msi.erb
+++ b/lib/chef/knife/bootstrap/windows-chef-client-msi.erb
@@ -91,18 +91,15 @@ goto Version6.2
   @set MACHINE_ARCH=<%= knife_config[:architecture] %>
 
   <% if knife_config[:architecture] == "x86_64" %>
-    IF "%PROCESSOR_ARCHITECTURE%"=="x86" (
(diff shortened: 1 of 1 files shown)


mwrock:
It would be great if you could provide more details in the PR description stating the problem you are trying to fix. Thanks!

mwrock:
:+1:

btm:
:+1:

charlesjohnson:
Closes #382 .

PR #385, opened 2016-08-23
Fixed win 2008 64bit ssh bootstrap command hanging 

dheerajd-msys:
During `knife bootstrap windows ssh` command for win 2008 64 bit, the command hangs at chef-client download through powershell command called through the bat file.

Setting InputFormat value to None in the powershell command resolved the issue and the bootstrap succeeded


mwrock:
:+1:

PR #388, opened 2016-08-25
fix #386 swallowing node_ssl_verify_mode value

### LINKED ISSUE #386: Problem with node-ssl-verify-mode none
I want to ignore ssl verification on my node. From my workstation I use `knife windows bootstrap winrm --node-ssl-verify-mode none` . I expect` client.rb` on the target to contain `ssl_verify_mode :verify_none`. It does not. If I add it to `client.rb` on the node manually after the bootstrap, things work much better.

Am I missing something? `windows_bootstrap_context.rb` sees to have code to handle the above but for me it's not working

mwrock:


smurawski:
:+1:

thommay:
:+1: LGTM

PR #389, opened 2016-08-26
Updated relay_command for using CommandExecutor class of winrm

Vasu1105:


Vasu1105:
@mwrock  & @btm This PR is ready for review.

lib/chef/knife/winrm_session.rb
```diff
@@ -49,27 +49,19 @@ def initialize(options)
       end
 
       def relay_command(command)
-        remote_id = @winrm_session.open_shell
-        command_id = @winrm_session.run_command(remote_id, command)
-        Chef::Log.debug("#{@host}[#{remote_id}] => :run_command[#{command}]")
-        session_result = get_output(remote_id, command_id)
-        @winrm_session.cleanup_command(remote_id, command_id)
-        Chef::Log.debug("#{@host}[#{remote_id}] => :command_cleanup[#{command}]")
+        command_exec = WinRM::CommandExecutor.new(@winrm_session)
```
mwrock: we can actually make this much simpler with the new winrm semantics:
```
@winrm_session.create_executor do |executor|
  executor.run_cmd(command) do |stdout, stderr|
    print_data(@host, stdout) if stdout
    print_data(@host, stderr, :red) if stderr
  end
end
```
Now we don't need the `get_output` method at all.

The logging is not necessary either because the underlying winrm gem produces almost identical logging. Just make sure to include this in the initializer:
```
@winrm_session.logger = Chef::Log
```

Vasu1105: @mwrock  I have updated the code as per comments plz review.

mwrock:
:+1:

PR #391, opened 2016-08-31
reset the config before runnuing tests in winrm_spec

mwrock:


mwrock:
Needs https://github.com/chef/mixlib-config/pull/42 to release to pass

PR #392, opened 2016-09-02
Use winrm v2 and allow users to pass a shell

mwrock:
This uses the just released winrm v2 gems and allows users to specify a winrm shell. This can be `cmd`, `powershell`, or `elevated`.  `cmd` is the default. `elevated` will run commands via a powershell based scheduled task.


rhass:
:+1:

PR #403, opened 2016-11-06
Added --bootstrap-url argument to add support for leveraging mixlib-i…

josh-hetland:
…nstall from an alternate source. updated documentation to reflect option. Changed the bootstrap_template_spec environment setup to reset the config options before each pass. Configuration changes were being retained between describe blocks causing tests to fail when the order they were run in changed.

Due to environmental constraints at my company, installers had to be sourced internally. I worked with our chef success team to set up an internal mirror of omnitruck and mixlib-install to be able to leverage install.sh and the installer determination of omnitruck. Use of the --msi-url option ended up being cumbersome while attempting to bootstrap several thousand windows servers of various versions and architectures, this change allowed me to leverage the install.ps1 that came along with the internal mixlib-install setup. I wanted to share it back in the event it was of benefit to anyone else.

Signed-off-by: josh-hetland <josh.hetland@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,19 @@ Currently (March 2016) the `stable` channel of omnibus (where downloads using th
 
 #### Using a custom install URL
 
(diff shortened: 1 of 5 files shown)


tas50:
I'm going to close this out since we've removed bootstrap functionality from this repository. Bootstrapping is now handled directly chef project repo.

PR #407, opened 2016-11-11
Added value for config_log_level and config_log_location

dheerajd-msys:
In progress

### DIFF
--- a/lib/chef/knife/core/windows_bootstrap_context.rb
+++ b/lib/chef/knife/core/windows_bootstrap_context.rb
@@ -79,12 +79,14 @@ cache_options     ({:path => "c:/chef/cache/checksums", :skip_expires => true})
           else
             client_rb << "# Using default node name (fqdn)\n"
           end
+
           if @chef_config[:config_log_level]
             client_rb << %Q{log_level :#{@chef_config[:config_log_level]}\n}
           else
             client_rb << "log_level        :info\n"
           end
-            client_rb << get_log_location()
+
+          client_rb << "log_location\t#{get_log_location}"
 
           # We configure :verify_api_cert only when it's overridden on the CLI
           # or when specified in the knife config.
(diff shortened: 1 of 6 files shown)


lib/chef/knife/core/windows_bootstrap_context.rb
```diff
@@ -70,22 +68,23 @@ def trusted_certs_script
...
             client_rb << %Q{node_name "#{@config[:chef_node_name]}"\n}
           else
             client_rb << "# Using default node name (fqdn)\n"
           end
+          if @chef_config[:config_log_level]
+            client_rb << %Q{log_level :#{@chef_config[:config_log_level]}\n}
+          else
+            client_rb << "log_level        :info\n"
+          end
+            client_rb << get_log_location()
```
btm: fix indenting on this line

lib/chef/knife/core/windows_bootstrap_context.rb
```diff
@@ -149,6 +148,24 @@ def config_content
...
+          elsif (@chef_config[:config_log_location].equal?(STDOUT))
+            return "log_location    STDOUT\n"
+          elsif (@chef_config[:config_log_location].equal?(STDERR))
+            return "log_location    STDERR\n"
+          elsif @chef_config[:config_log_location].nil? || @chef_config[:config_log_location].empty?
+            return "log_location    STDOUT\n"
+          elsif @chef_config[:config_log_location]
+            return %Q{log_location  "#{@chef_config[:config_log_location]}"\n}
+          else
+            return "log_location    STDOUT\n"
```
btm: being the last statement in the method, you shouldn't need to use `return` anywhere here. Only one conditional be matched and we would return the string.

Given that you always return "log_location<tab>", I would have this method only return the value portion of that line and around line 87 use something like:

```
client_rb << "log_location\t#{get_log_location}"
```

btm:
This is the knife-windows duplication of https://github.com/chef/chef/pull/5502.

PR #409, opened 2016-11-16
Fix trusted_cert copy script generation on windows

### LINKED ISSUE #404: bootstrap windows does not copy contents of trusted_certs from a Windows workstation
When bootstrapping a node from a Windows workstation, using knife-windows 1.7.0, the contents of the workstation's `trusted_certs` folder are not copied to the `trusted_certs` folder on the node due to a bug in the `trusted_certs_content` method. I have verified that this does work properly on a Mac. The issue is that [Dir.glob does not take backslashes](https://stackoverflow.com/questions/20667542/dir-glob-not-returning-any-files/20668433#20668433), yet `PathHelper.escape_glob(@chef_config[:trusted_certs_dir])` returns double backslashes on a Windows machine.

The existing `trusted_certs_content` method:
https://github.com/chef/knife-windows/blob/master/lib/chef/knife/core/windows_bootstrap_context.rb

```
        # Returns a string for copying the trusted certificates on the workstation to the system being bootstrapped
        # This string should contain both the commands necessary to both create the files, as well as their content
        def trusted_certs_content
          content = ""
          if @chef_config[:trusted_certs_dir]
            Dir.glob(File.join(PathHelper.escape_glob(@chef_config[:trusted_certs_dir]), "*.{crt,pem}")).each do |cert|
              content << "> #{bootstrap_directory}/trusted_certs/#{File.basename(cert)} (\n" +
                         escape_and_echo(IO.read(File.expand_path(cert))) + "\n)\n"
            end
          end
          content
        end
```

The fix is simple. Just add `.gsub('\\','/')` so we do this instead:

```
        # Returns a string for copying the trusted certificates on the workstation to the system being bootstrapped
        # This string should contain both the commands necessary to both create the files, as well as their content
        def trusted_certs_content
          content = ""
          if @chef_config[:trusted_certs_dir]
            Dir.glob(File.join(PathHelper.escape_glob(@chef_config[:trusted_certs_dir]), "*.{crt,pem}").gsub('\\','/')).each do |cert|
              puts "cert: '#{File.basename(cert)}'"
              content << "> #{bootstrap_directory}/trusted_certs/#{File.basename(cert)} (\n" +
                         escape_and_echo(IO.read(File.expand_path(cert))) + "\n)\n"
            end
          end
          content
        end
```

coderjoe:
Trusted certs are not being copied because Dir.glob
can't handle the windows path separators being returned
by PathHelper.escape_glob. This solves the problem by
replacing those separators with unix style path separators.

Fixes #404

Signed-off-by: Joseph Bauser <coderjoe@coderjoe.net>


coderjoe:
I'll address the failing travis build shortly.

mwrock:
Cool! So I'm taking a closer look at this and I think the right fix is to use `escape_glob_dir` which we should "steal" from https://github.com/chef/chef/blob/master/chef-config/lib/chef-config/path_helper.rb#L166. Note the comments there about `escape_glob`.

Its unfortunate that we have to "copy/paste" here but we do that so that knife-windows can support chef 11. We try to support `Major Version - 1`.

I really like that you added tests here. You could either add that method in or I can add a commit to this PR either works for me.

coderjoe:
I'm on the road right now @mwrock I can make changes when I get home in an hour or so or you can go ahead now if you're already spelunking. Either way works for me! :+1:

mwrock:
Sounds good I'll just commit on top of yours. Thanks!!

coderjoe:
If you pick it up, be aware that the travis tests may be a bit too naive. The ordering will likely be wrong on at least one of the platforms. It should probably be testing to make sure the script contains the certs, not necessarily rely on the fact that the script is line for line identical anyway.

mwrock:
Yeah it looks like both windows and linux hosts will order the certs in the opposite order. Thats just greaaat. Will fix it up in a bit.

coderjoe:
I'm back! @mwrock if you'd like I can merge your work back into my branch and fix up those tests. For the sake of simplicity I'll presume you've already got it covered unless you say otherwise. Cheers!

mwrock:
Perfect if you can/want! Otherwise I will do it when I am back.

coderjoe:
Ok, I'll give it a go! Thanks!

coderjoe:
@mwrock tests are green. Do me a favor and give it a look and, presuming it passes muster, let me know if you'd like me to squash the commits down prior to merging. Thanks for your help!

PR #416, opened 2017-01-24
Add concurrency support to knife-windows

### LINKED ISSUE #348: knife winrm command parallelism
I have a command we run in order to run chef-client on all our windows machines in an environment:

`knife winrm "chef_environment:?? AND platform:windows" "chef-client" -x ?? -P ??`

When using Chefdk 0.9.0 (with knife-windows 0.8.6) this command executes chef-client on all those machine in parallel, so only taking as long as the longest chef-client run.  However using Chefdk 0.10.0 (with knife-windows >= 1.0.0) this command executes chef-client one after another, taking much much longer to run.

Am I missing something?  Was parallelism in knife wirm removed?

Thanks

tduffield:
Until the native WinRM libraries support concurrent connections, add simple threading to handle parallel connections. 

Fixes #348

Signed-off-by: Tom Duffield <tom@chef.io>

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,8 @@ The `winrm` subcommand allows you to invoke commands in parallel on a subset of 
 
     knife winrm "role:web" "net stats srv" -x Administrator -P 'super_secret_password'
 
+_Please note that to run a single command against multiple nodes, each node must share the same username and password._
+
 Or force a chef run:
 
     knife winrm "myserver.myorganization.net" "chef-client -c c:/chef/client.rb" -m -x Administrator -P "super_secret_password"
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -378,6 +380,33 @@ local to the target Windows system.
...
+$logfile="C:\\Program Files\\Amazon\\Ec2ConfigService\\Logs\\kitchen-ec2.log"
+# Allow script execution
+Set-ExecutionPolicy -ExecutionPolicy RemoteSigned -Force
+#PS Remoting and & winrm.cmd basic config
+Enable-PSRemoting -Force -SkipNetworkProfileCheck
+& winrm.cmd set winrm/config '@{MaxTimeoutms="1800000"}' >> $logfile
+& winrm.cmd set winrm/config/winrs '@{MaxMemoryPerShellMB="1024"}' >> $logfile
+& winrm.cmd set winrm/config/winrs '@{MaxShellsPerUser="50"}' >> $logfile
+#Server settings - support username/password login
+& winrm.cmd set winrm/config/service/auth '@{Basic="true"}' >> $logfile
```
smurawski: With the current capabilities in the WinRM gem, we don't need the Basic='true' and AllowUnencrypted settings anymore.

I know I pointed you at an example from kitchen-ec2, but we've got to fix that there too. :)  Sorry.

lib/chef/knife/winrm.rb
```diff
@@ -49,7 +49,12 @@ def run
         STDOUT.sync = STDERR.sync = true
 
         configure_session
-        execute_remote_command
+        exit_status = execute_remote_command
+        if exit_status != 0
+          exit exit_status
```
smurawski: If this happens in a parallel bootstrap, I believe this would interrupt other, ongoing bootstraps.

tduffield: Should the thread join not block the parent thread before it returns up through execute_remote_command?

smurawski: You're right, this is cool as is.

aqueeb:
Thanks a lot! You guys are awesome!

PR #421, opened 2017-02-23
fix version comparison operations

lamont-granquist:
these don't work right, eg.:

Chef::VERSION.split('.')[0..1].join('.').to_f >= 12.5

12.130 is not >= 12.500

but chef 12.13 certainly is >= 12.5


lamont-granquist:
there's some busted specs, gotta figure out what was actually intended because they clearly weren't working correctly before.

PR #434, opened 2017-07-19
Expand testing to match ruby versions we support

tas50:
Remove the ability to test on Chef 11. We're not doing this in Travis anymore
Fix tests when running on Chef 13
Properly compare versions so the spec exclusion logic works
Expand the Ruby versions we test on

Signed-off-by: Tim Smith <tsmith@chef.io>


tas50:
This is going to require more work. The specs need cleanup for modern Chef 12 releases.

tas50:
@btm this should get things green so we can move forward on the other PRs

Issue #412, opened 2016-11-18
Bootstrapping Windows 2012 ends with \Microsoft was unexpected at this time


kriston:
Bootstrapping Windows 2012 ends with \Microsoft was unexpected at this time.  This happens with 1.7.1 and 1.7.0.

This is the output when it attempts this step in the bootstrap:

`
DEBUG: [WinRM] opening remote shell on http://10.0.0.1:5985/wsman
DEBUG: [WinRM] remote shell created with shell_id: B0FD312C-9CC0-44E6-B091-C7B4C28139B6
[...]

Issue #318, opened 2015-12-02
Windows bootstrap should run in anElevated session under System account


Annih:

jvogt:
Just curious if anyone has a workaround for this, short of running chef client while logged into the machine?

gitdave225:
We have our chef client installed as a service. It checks back with the master server every half an hour.

mwrock:
once https://github.com/WinRb/WinRM/pull/191 is merged, I'd like to do some winrm refactoring here to atleast add a flag to knife winrm that will run commands through winrm-elevated.

SolverDude:
@mwrock SWEET!  Is this imminent?  Can I help?

mwrock:
Just waiting on a green light  from @sneal. It's quite a bit to review.

jptoto:
@mwrock did this or something like it ever get merged? Sure would be handy to have a flag for knife bootstrap to run as elevated.

mwrock:
you can now run ad hoc knife winrm commands elevated using `--winrm-shell elevated` but bootstrap continues on in cmd.exe because we share the same scripts for knife bootstrap ssh.

jptoto:
That'll work for now. Thanks @mwrock much appreciated 👍

PR #435, opened 2017-09-26
Run Windows bootstrap under SYSTEM account triggered by scheduled task

RobbertJanSW:
* Better error handling (in case MSI download fails / MSI install fails, etc)
* Dump MSI install log in case of install error
* Smoother install (wait for the MSI install to finish, continue afterwards instead of just waiting 300 sec)
* Runs under SYSTEM account (scheduled task) instead of logged in user account and remote WINRM restrictions
* Added debug option (which skips the `@echo off` in the bootstrap.cmd script)


RobbertJanSW:
By the way this possibly fixes #412

RobbertJanSW:
And of coarse #318

tas50:
I'm going to close this out since we've removed bootstrap functionality from this repository. Bootstrapping is now handled directly chef project repo.

PR #444, opened 2018-03-01
Change the %PATH% env to use windows default values

teknofire:
Fixes issue when bootstrapping windows systems failing with the message: `The input line is too long.`.  

This is triggered by a "very" long `%PATH%` environment variable getting expanded at execution and causing the command being execute to become longer than the allowed size. 

We only need enough of the path set to be to run chef-client and doesn't persist to the system.

Signed-off-by: Will Fisher <wfisher@chef.io>

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -79,6 +79,10 @@ def chef_lt_14?
   Gem::Version.new(Chef::VERSION) < Gem::Version.new('14')
 end
 
+def chef_gte_14?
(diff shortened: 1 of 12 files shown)


spec/spec_helper.rb
```diff
@@ -79,6 +79,10 @@ def chef_lt_14?
   Gem::Version.new(Chef::VERSION) < Gem::Version.new('14')
 end
 
+def chef_gte_14?
```
tas50: YES!

tas50:
Thanks for the spec helper fix as well

PR #453, opened 2018-04-18
Fix the specs by ignoring DEPRECATED & unimplemented options

NAshwini:
Fix the specs by ignoring DEPRECATED options "distro" & "template_file" & by ignoring unimplemented option in knife-windows i.e "bootstrap_preinstall_command".
Fix for issue https://github.com/chef/chef/issues/7125
Signed-off-by: NAshwini <ashwini.nehate@msystechnologies.com>

### DIFF
--- a/spec/unit/knife/bootstrap_options_spec.rb
+++ b/spec/unit/knife/bootstrap_options_spec.rb
@@ -101,6 +101,9 @@ expected: #{expected}
       :ssh_gateway_identity,
       :bootstrap_proxy_user,
       :bootstrap_proxy_pass,
+      :bootstrap_preinstall_command,
+      :distro, # Deprecated - remove this when the flag is removed.
+      :template_file, # Deprecated - remove this when the flag is removed.
     ]}
 
     # win_ignore: Options in windows that aren't relevant to core.
@@ -120,6 +123,8 @@ expected: #{expected}
       :winrm_codepage,
       :concurrency,
       :winrm_shell,
+      :distro, # Deprecated - remove this when the flag is removed.
+      :template_file, # Deprecated - remove this when the flag is removed.
     ] }
 
     include_examples 'compare_options'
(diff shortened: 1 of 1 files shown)


NAshwini:
Hi @btm, the PR is failing for mismatch of version for `mixlib-log` gem file which is unrelated to this PR commits.

spec/unit/knife/bootstrap_options_spec.rb
```diff
@@ -120,6 +123,8 @@ def compare_property(sym)
       :winrm_codepage,
       :concurrency,
       :winrm_shell,
+      :distro, # Deprecated - remove this when the flag is removed.
+      :template_file, # Deprecated - remove this when the flag is removed.
```
btm: these are just supposed to be the options that are unique to knife-windows. I don't think you want these two lines.

spec/unit/knife/bootstrap_options_spec.rb
```diff
@@ -148,12 +153,17 @@ def compare_property(sym)
...
     ]}
     # win_ignore: Options in windows that aren't relevant to core.
     let(:win_ignore) { [
       :auth_timeout,
       :install_as_service,
       :host_key_verification,  # Deprecated - remove this when the flag is removed.
+      :distro, # Deprecated - remove this when the flag is removed.
+      :template_file, # Deprecated - remove this when the flag is removed.
```
btm: i think these two aren't supposed to be here? options shouldn't be in both the `ref_ignore` and `win_ignore`.

NAshwini: Hi @btm, as per my first commit `distro` & `template_file` options was only added in `win_ignore` & locally it was working fine for all 10 example successfully on `chef 14` & `chef 13` as well. But travis was getting fail on `chef 13` for example https://github.com/MsysTechnologiesllc/knife-windows/blob/d3130b14110e2276c6bcd1b8db8faf2b5a71081e/spec/unit/knife/bootstrap_options_spec.rb#L47, I was not able reprodcue travis issue locally so I have added it in `ref_ignore` too to pass the travis.

PR #454, opened 2018-05-28
Clear pem before writing

RobbertJanSW:


tas50:
@RobbertJanSW Looks like this got hit with a rebase that merged in a bunch of other changes. If you're still interested in making this change to knife-windows can you open up a new PR that has just the single change?

Issue #449, opened 2018-03-16
Bootstrap template (`-t`) does not work


spuder:
Chef Development Kit Version: 1.6.11
chef-client version: 12.21.26
delivery version: master (73ebb72a6c42b3d2ff5370c476be800fee7e5427)
berks version: 5.6.4
kitchen version: 1.16.0
inspec version: 1.25.1

knife-windows (1.9.0)

If you use the `-t` option for a bootstrap template, the bootstrap option is completely ignored. 
If you use the `--bootstrap-template` option, the bootstrap option works as expected. 

# Steps to reproduce

1. Put dummy data in a text file

c:\users\foobar\.chef\bootstrap\foo.erb

```
@rem stuff
@echo ######################
@echo ######################
```

2. boostrap using the `-t` option

--bootstrap-template gives expected result of garbage template (#####)
```
knife bootstrap windows winrm 192.0.2.0 --bootstrap-template 'c:/users/foobar/.chef/bootstrap/foobar.erb'
Creating new client for foobar
Creating new node for foobar

Waiting for remote response before bootstrap.192.0.2.0 .
192.0.2.0 Response received.
Remote node responded after 0.53 minutes.
Bootstrapping Chef on 192.0.2.0
192.0.2.0 Rendering "C:\Users\vagrant\AppData\Local\Temp\bootstr
4.bat" chunk 1
192.0.2.0 ##########################################################################
192.0.2.0 ##########################################################################
```

-t gives unexpected result of using default template
```
knife bootstrap windows winrm 192.0.2.0 -t 'c:/users/foobar/.chef/bootstrap/foobar.erb'
Creating new client for foobar
Creating new node for foobar

Waiting for remote response before bootstrap.192.0.2.0 .
192.0.2.0 Response received.
Remote node responded after 0.01 minutes.
Bootstrapping Chef on 192.0.2.0
192.0.2.0 Rendering "C:\Users\vagrant\AppData\Local\Temp\bootstr
6.bat" chunk 1
192.0.2.0 Rendering "C:\Users\vagrant\AppData\Local\Temp\bootstr
6.bat" chunk 2
192.0.2.0 Rendering "C:\Users\vagrant\AppData\Local\Temp\bootstr
6.bat" chunk 3
192.0.2.0 Rendering "C:\Users\vagrant\AppData\Local\Temp\bootstr
6.bat" chunk 4
192.0.2.0 Rendering "C:\Users\vagrant\AppData\Local\Temp\bootstr
[...]

PR #457, opened 2018-06-19
[MSYS-841]fix bootstrap template short name

dheerajd-msys:
**Description**
For command `knife bootstrap windows winrm` : 
If we use the `-t` option for a bootstrap template, the bootstrap option is completely ignored.
If we use the `--bootstrap-template` option, the bootstrap option works as expected. 
This happens because `-t` was being used for `--winrm-transport` as well as  `--bootstrap-template`

**Issue Resolved**
https://github.com/chef/knife-windows/issues/449

### DIFF
--- a/lib/chef/knife/bootstrap_windows_base.rb
+++ b/lib/chef/knife/bootstrap_windows_base.rb
@@ -74,21 +74,21 @@ class Chef
           option :distro,
             :short => "-d DISTRO",
             :long => "--distro DISTRO",
-            :description => "Bootstrap a distro using a template. [DEPRECATED] Use -t / --bootstrap-template option instead.",
+            :description => "Bootstrap a distro using a template. [DEPRECATED] Use -b / --bootstrap-template option instead.",
             :proc        => Proc.new { |v|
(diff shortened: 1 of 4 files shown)


dheerajd-msys:
@PrajaktaPurohit, @btm  Could you please review and merge?

Thanks

lib/chef/knife/bootstrap_windows_base.rb
```diff
@@ -74,21 +74,21 @@ def self.included(includer)
...
+            :description => "Bootstrap a distro using a template. [DEPRECATED] Use -b / --bootstrap-template option instead.",
             :proc        => Proc.new { |v|
               Chef::Log.warn("[DEPRECATED] -d / --distro option is deprecated. Use --bootstrap-template option instead.")
               v
             }
 
           option :bootstrap_template,
-            :short => "-t TEMPLATE",
-            :long => "--bootstrap-template TEMPLATE",
+            :short => "-b BOOTSTRAP_TEMPLATE",
```
rhass: Personally, I would rather we change the short option flag for the winrm transport so that the common parameters between knife bootrstrap, and knife bootstrap winrm remain the same. Other than this one nit, everything looks good.

PR #458, opened 2018-07-03
[MSYS-850] enable expeditor

dheerajd-msys:


VERSION
```diff
@@ -0,0 +1 @@
+1.9.1
```
btm: This should be the next version, 1.9.2

https://github.com/chef/knife-ec2/pull/525#issuecomment-392798723

btm:
I added the Expeditor related Github labels, verified the repository permissions are set up for Expeditor, and verified it can release the Rubygem, so we just need a couple little changes here and we can merge this.

PR #460, opened 2018-10-15
Fix to add client.d support.

vijaymmali1990:
Signed-off-by: vijaymmali1990 <vijay.mali@msystechnologies.com>
There was issue with creating sub directories and files in client.d dir in [Existing PR](https://github.com/chef/knife-windows/pull/443) . This solution will fix the issue.


btm:
Combined this into #461

Issue #469, opened 2019-01-17
Remove the deprecated identity_file option


tas50:
This option has been deprecated for many years and was replaced with ssh_identity_file. Make sure we're using ssh_identity_file everywhere in knife-windows instead of identify_file.

dheerajd-msys:
As we agreed upon to keep `--identity-file` option until `chef v14` goes EOL. Other deprecated options are removed in the PR [chef/knife-windows#470](https://github.com/chef/knife-windows/pull/470)

PR #470, opened 2019-01-18
Removed deprecated host_key_verification, distro and template_file options

dheerajd-msys:

## Descriptions

- Removed deprecated `--host-key-verification`, `--distro` & `--template-file` option.
- `--host-key-verify` is used now instead of `--host-key-verification`
- `--bootstrap-template-file` is used now instead of  `--distro` & `--template-file`

## Resolved Issue
Fixes https://github.com/chef/knife-windows/issues/469

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,6 @@ gemfile: ci.gemfile
 env:
   - CHEF_VERSION="master"
   - CHEF_VERSION="~> 14.0"
-  - CHEF_VERSION="~> 13.0"
 
 matrix:
   exclude:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -9,7 +9,7 @@ environment:
 
   matrix:
     - ruby_version: "24"
-      chef_version: "~> 13.0"
+      chef_version: "~> 14.0"
 
     - ruby_version: "25"
       chef_version: "master"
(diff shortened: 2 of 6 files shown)


tas50:
Agree with @rhass here. Let's make sure we nuke every mention of this old option including our tests.

dheerajd-msys:
Thanks @tas50 and @rhass I will do that.

dheerajd-msys:
We may need to update travis and appveyor to get all this green. Travis and AppVeyor is failing for `Environment: ruby_version=24, chef_version=~>13.0`. 

@rhass & @tas50 Could you please take a look?

Vasu1105:
@dheerajd-msys : We should remove --template-file and --distro option also in this same PR.

tas50:
@dheerajd-msys looks like travis is failing on this spec that's still using the distro flag: https://github.com/chef/knife-windows/blob/master/spec/unit/knife/bootstrap_windows_winrm_spec.rb#L48

tas50:
As part of this PR you can remove Chef 13 testing since we'll require Chef 14 or later. We probably want to note that in the readme. You can also change the minimum required ruby version to 2.4 in the gemspec since that's the same as what Chef 14 required.

dheerajd-msys:
@tas50 Problem here is that even if we remove `chef v13` and use `chef v14` it will still not make travis happy. Because `--identity-file` option is not removed yet from `chef v14` as you can see here https://github.com/chef/chef/blob/v14.9.13/lib/chef/knife/bootstrap.rb#L81.

tas50:
We will probably need to keep identity_file for now then.

PR #471, opened 2019-01-23
Using chef/chef path_helper and removed the knife-windows path_helper

### LINKED ISSUE #467: Use the Chef path_helper instead of the vendored version
knife-windows currently has it's own path_helper library that was included because the path_helper in Chef 11 did not include the home method. We can assume everyone is on Chef 12 or later at this point and switch back to Chef's own helper.

Vasu1105:
Signed-off-by: vasu1105 <vasundhara.jagdale@msystechnologies.com>

As mentioned in #467 knife-windows currently has it's own path_helper library that was included because the path_helper in Chef 11 did not include the home method. We can assume everyone is on Chef 12 or later at this point and switch back to Chef's own helper

Fixed #467


tas50:
Thanks @Vasu1105

PR #477, opened 2019-04-11
Prep branch for knife-windows v3

btm:
knife-windows v3 will support only chef >= 15, which contains support for bootstrapping windows.

I've already created a `2-stable` branch.


appveyor.yml
```diff
@@ -8,8 +8,6 @@ environment:
   bundle_gemfile: ci.gemfile
 
   matrix:
-    - ruby_version: "24"
-      chef_version: "~> 14.0"
     - ruby_version: "25"
```
tas50: We should get ruby 26 added here since they have it

.travis.yml
```diff
@@ -4,7 +4,6 @@ dist: xenial
 sudo: false
 
 rvm:
-  - 2.4.5
   - 2.5.3
   - 2.6.0
```
tas50: These can get bumped to the new patch releases

.travis.yml
```diff
@@ -4,7 +4,6 @@ dist: xenial
 sudo: false
```
tas50: Sudo config can go away now that they don’t use it for anything

Issue #479, opened 2019-05-03
Chef 15: uninitialized constant Chef::Knife::Bootstrap::ClientBuilder


btm:
# Version:

knife-windows 3.0.0 and chef 15.0.256

# Environment:

ubuntu 18.04.02

# Scenario:

test knife-windows v3

# Steps to Reproduce:

```
# install the latest Chef omnibus package from current
# from knife-windows source:
rake build
sudo /opt/chef/embedded/bin/gem install pkg/knife-windows-3.0.0.gem
knife bootstrap windows winrm
```

# Expected Result:

displays deprecation message from #478 

# Actual Result:

```
$ knife bootstrap windows  winrm
/opt/chef/embedded/lib/ruby/gems/2.6.0/gems/chef-15.0.256/lib/chef/knife/bootstrap.rb:28: warning: already initialized constant Chef::Knife::Bootstrap::SUPPORTED_CONNECTION_PROTOCOLS
/opt/chef/embedded/lib/ruby/gems/2.6.0/gems/chef-15.0.256/lib/chef/knife/bootstrap.rb:28: warning: previous definition of SUPPORTED_CONNECTION_PROTOCOLS was here
/opt/chef/embedded/lib/ruby/gems/2.6.0/gems/chef-15.0.256/lib/chef/knife/bootstrap.rb:29: warning: already initialized constant Chef::Knife::Bootstrap::WINRM_AUTH_PROTOCOL_LIST
/opt/chef/embedded/lib/ruby/gems/2.6.0/gems/chef-15.0.256/lib/chef/knife/bootstrap.rb:29: warning: previous definition of WINRM_AUTH_PROTOCOL_LIST was here
/opt/chef/embedded/lib/ruby/gems/2.6.0/gems/chef-15.0.256/lib/chef/knife/bootstrap.rb:339: warning: already initialized constant Chef::Knife::Bootstrap::DEPRECATED_FLAGS
/opt/chef/embedded/lib/ruby/gems/2.6.0/gems/chef-15.0.256/lib/chef/knife/bootstrap.rb:339: warning: previous definition of DEPRECATED_FLAGS was here
Traceback (most recent call last):
	6: from /opt/chef/bin/knife:148:in `<main>'
	5: from /opt/chef/bin/knife:148:in `load'
	4: from /opt/chef/embedded/lib/ruby/gems/2.6.0/gems/chef-15.0.256/bin/knife:24:in `<top (required)>'
[...]

PR #480, opened 2019-05-06
Load bootstrap dependency

vsingh-msys:
Signed-off-by: Vivek Singh <vivek.singh@msystechnologies.com>

### Description

Load dependency required files for knife bootstrap class. 

### Issues Resolved
https://github.com/chef/knife-windows/issues/479

### DIFF
--- a/lib/chef/knife/bootstrap_windows_base.rb
+++ b/lib/chef/knife/bootstrap_windows_base.rb
@@ -16,6 +16,8 @@
 # limitations under the License.
 #
 
+require 'chef/knife'
+require 'chef/knife/bootstrap'
 require 'chef/encrypted_data_bag_item'
 require 'chef/knife/knife_windows_base'
 require 'chef/util/path_helper'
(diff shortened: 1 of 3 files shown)


marcparadise:
This change makes sense - for Chef 15 those requires were moved into `deps` for bootstrap.rb

However, I'm not clear how we were able to do away with the require for `chef/knife/bootstrap` - that's still needed to reference the class and its `load_deps`.  Is bootstrap already being required in by knife itself?

vsingh-msys:
Hi @marcparadise ,
> However, I'm not clear how we were able to do away with the require for chef/knife/bootstrap - that's still needed to reference the class and its load_deps. Is bootstrap already being required in by knife itself?

No, it's included by knife by default.
But somehow it's available at that moment and if we add require 
it's loaded multiple times to solve this I removed those require classes.

BTW to be safe side going to add remove requires files.

Thanks

driesgroblerw:
With which version of the Chef Infra client will/has this fix been released?

Issue #445, opened 2018-03-05
knife winrm stops when there is one host which is not responding


liivats:
If you using command knife _winrm **_"platform_family:windows" "hostname"_**
Then it fails when reaches the host which is not responding. To compare knife ssh, then knife ssh works as expected. It resumes when there is non-responsive host. 

Please fix it as it is really important to be able to proceed with hosts which running.

PS C:\windows\system32> **_knife winrm "platform_family:windows" "hostname" -x 'username'_**
Enter your password:
**hostname1 hostname1**
ERROR: Network Error: A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond. - A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond. - connect(2) for **"hostname2" port 5985 (hostname2:5985)**
Check your knife configuration and network settings
PS C:\windows\system32>

bish0polis:
This looks like #402 .

Vasu1105:
@liivats sure we will look into this on priority.

Vasu1105:
I have identified that to connect to multiple servers using knife winrm command you need to also provide an `--concurrency The number of allowed concurrent connection` option along with your `--manual-list                QUERY is a space separated list of servers` option

The command which I verified is

```knife winrm -C 3 -m "<ip-address-1> <ip-address-2> <ip-address-2>" 'hostname' -x "azure" -P "azure@12345678" -c ~/workspace/chef-repo/.chef/knife.rb ```

and the result which I got is it did listed the hostname for the servers which are responding and then at the end the end I got timeout error for the server which I purposely stopped 
```
<ip-address-1> vj-msys-891
<ip-address-3>  vj-msys-891-is2

/home/msys/workspace/knife-windows/bundle/ruby/2.5.0/gems/httpclient-2.8.3/lib/httpclient/session.rb:611:in `initialize': execution expired (HTTPClient::ConnectTimeoutError)
ERROR: HTTPClient::ConnectTimeoutError: execution expired
```
@btm do you think that based on the number of servers that are coming as input from user we should decided the concurrency and set it dynamically rather that expecting that to be provided by user?

NAshwini:
Closing this issue as its been resolved by PR https://github.com/chef/knife-windows/pull/484

Issue #402, opened 2016-11-02
WinRM Connection fails when one node in list is down


mtharpe:
When running knife winrm "name:*somequery*" 'some command' and there is a return from the chef server that has a server that is not available the whole command fails. 

/opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/httpclient-2.8.2.4/lib/httpclient/session.rb:604:in `initialize': execution expired (HTTPClient::ConnectTimeoutError)
	from /opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/httpclient-2.8.2.4/lib/httpclient/session.rb:604:in `new'
	from /opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/httpclient-2.8.2.4/lib/httpclient/session.rb:604:in `create_socket'
	from /opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/httpclient-2.8.2.4/lib/httpclient/session.rb:749:in `block in connect'
	from /opt/chefdk/embedded/lib/ruby/2.3.0/timeout.rb:101:in `timeout'
	from /opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/httpclient-2.8.2.4/lib/httpclient/session.rb:742:in `connect'
	from /opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/httpclient-2.8.2.4/lib/httpclient/session.rb:504:in `query'
	from /opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/httpclient-2.8.2.4/lib/httpclient/session.rb:174:in `query'
	from /opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/httpclient-2.8.2.4/lib/httpclient.rb:1240:in `do_get_block'
	from /opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/httpclient-2.8.2.4/lib/httpclient.rb:1017:in `block in do_request'
	from /opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/httpclient-2.8.2.4/lib/httpclient.rb:1131:in `protect_keep_alive_disconnected'
	from /opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/httpclient-2.8.2.4/lib/httpclient.rb:1012:in `do_request'
[...]

PR #484, opened 2019-08-29
Fix multiple session issue of concurrency flag

NAshwini:
Signed-off-by: NAshwini <ashwini.nehate@msystechnologies.com>

### Description
Multiple servers can connect now using knife winrm command with and without providing an `--concurrency The number of allowed concurrent connection` option along with your `--manual-list QUERY is a space separated list of servers` option

### Issues Resolved
https://github.com/chef/knife-windows/issues/445
https://github.com/chef/knife-windows/issues/402

### Check List

- [*] New functionality includes tests
- [*] All tests pass
- [*] All commits have been signed-off for the Developer Certificate of Origin. See <https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco>
- [ ] PR title is a worthy inclusion in the CHANGELOG

### DIFF
--- a/lib/chef/knife/winrm_knife_base.rb
+++ b/lib/chef/knife/winrm_knife_base.rb
@@ -118,15 +118,16 @@ class Chef
           def relay_winrm_command(command)
             Chef::Log.debug(command)
             @session_results = []
-
             queue = Queue.new
             @winrm_sessions.each { |s| queue << s }
-            number_of_session = config[:concurrency] ? locate_config_value(:concurrency) : @name_args[0].split(" ").count || queue.size
-            # These nils will kill the Threads once no more sessions are left
-            number_of_session.times { queue << nil }
+            num_sessions = locate_config_value(:concurrency)
+            num_targets = @winrm_sessions.length
+            num_sessions = (num_sessions.nil? || num_sessions == 0) ? num_targets : [num_sessions, num_targets].min
 
+            # These nils will kill the Threads once no more sessions are left
+            num_sessions.times { queue << nil }
             threads = []
-            number_of_session.times do
+            num_sessions.times do
               threads << Thread.new do
                 while session = queue.pop
                   run_command_in_thread(session, command)
--- a/lib/chef/knife/winrm_knife_base.rb
+++ b/lib/chef/knife/winrm_knife_base.rb
@@ -121,11 +121,12 @@ class Chef
 
             queue = Queue.new
             @winrm_sessions.each { |s| queue << s }
+            number_of_session = config[:concurrency] ? locate_config_value(:concurrency) : @name_args[0].split(" ").count || queue.size
             # These nils will kill the Threads once no more sessions are left
-            locate_config_value(:concurrency).times { queue << nil }
+            number_of_session.times { queue << nil }
 
             threads = []
-            locate_config_value(:concurrency).times do
+            number_of_session.times do
               threads << Thread.new do
                 while session = queue.pop
                   run_command_in_thread(session, command)
--- a/lib/chef/knife/winrm_shared_options.rb
+++ b/lib/chef/knife/winrm_shared_options.rb
@@ -43,7 +43,7 @@ class Chef
           option :concurrency,
             short: "-C NUM",
             long: "--concurrency NUM",
-            description: "The number of allowed concurrent connections",
+            description: "The number of allowed concurrent connections - default is number of hosts",
             proc: lambda { |o| o.to_i }
         end
       end
--- a/lib/chef/knife/winrm_shared_options.rb
+++ b/lib/chef/knife/winrm_shared_options.rb
@@ -44,7 +44,6 @@ class Chef
             short: "-C NUM",
             long: "--concurrency NUM",
             description: "The number of allowed concurrent connections",
-            default: 1,
             proc: lambda { |o| o.to_i }
         end
       end
--- a/spec/unit/knife/winrm_spec.rb
+++ b/spec/unit/knife/winrm_spec.rb
@@ -447,6 +447,22 @@ describe Chef::Knife::Winrm do
       expect { @winrm.run_with_pretty_exceptions }.to raise_error(SystemExit)
     end
 
+    context "impact of concurrency value" do
+      before(:each) do
+        allow(Chef::Knife::WinrmSession).to receive(:new).and_return(session)
+        @winrms = Chef::Knife::Winrm.new(['-C', '3', '-m', 'localhost', '-x', 'testuser', '-P', 'testpassword', '--winrm-authentication-protocol', 'basic', 'echo helloworld'])
+      end
+      it "create sessions to equal number of concurrent value if concurrent option is given" do
+        @winrms.config[:concurrency] = 1
+        allow(@winrms).to receive(:relay_winrm_command).and_return(true)
+        expect { @winrms.run_with_pretty_exceptions }.not_to raise_error
+      end
+      it "create sessions to equal number of IP's if concurrent option is not given" do
+        allow(@winrm).to receive(:relay_winrm_command).and_return(true)
+        expect { @winrm.run_with_pretty_exceptions }.not_to raise_error
+      end
+    end
+
     context "when winrm_authentication_protocol specified" do
       before do
         Chef::Config[:knife] = { winrm_transport: "plaintext" }
--- a/spec/unit/knife/winrm_spec.rb
+++ b/spec/unit/knife/winrm_spec.rb
@@ -447,22 +447,6 @@ describe Chef::Knife::Winrm do
       expect { @winrm.run_with_pretty_exceptions }.to raise_error(SystemExit)
     end
 
-    context "impact of concurrency value" do
-      before(:each) do
-        allow(Chef::Knife::WinrmSession).to receive(:new).and_return(session)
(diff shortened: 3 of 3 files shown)


lib/chef/knife/winrm_knife_base.rb
```diff
@@ -122,11 +122,12 @@ def relay_winrm_command(command)
 
             queue = Queue.new
             @winrm_sessions.each { |s| queue << s }
+            number_of_session = config[:concurrency] ? locate_config_value(:concurrency) : @name_args[0].split(" ").count || queue.size
```
Nimesh-Msys: `@name_args[0].split(" ").count || queue.size`
As count would always return a number, `queue.size` would be non-reachable here.
Could you please confirm if we need a `zero?` check with the count or may be an empty check with the array here.

lib/chef/knife/winrm_knife_base.rb
```diff
@@ -122,11 +122,12 @@ def relay_winrm_command(command)
...
             # These nils will kill the Threads once no more sessions are left
-            locate_config_value(:concurrency).times { queue << nil }
+            number_of_session.times { queue << nil }
```
Nimesh-Msys: Array's default would anyway be nil, may be we can tweak the upcoming iteration logic.
Still if we want this, `queue += [nil] * number_of_sessions` would be a faster and nicer choice.
A sample Benchmark comparison with 30000 iterations:
Current: 6.8137
Suggested: 0.3112

lib/chef/knife/winrm_shared_options.rb
```diff
@@ -44,7 +44,6 @@ def self.included(includer)
             :short => "-C NUM",
             :long => "--concurrency NUM",
```
Nimesh-Msys: Since we are touching this, can we use new hash formats ?

spec/unit/knife/winrm_spec.rb
```diff
@@ -443,6 +443,22 @@
       expect { @winrm.run_with_pretty_exceptions }.to raise_error(SystemExit)
     end
 
+    context "impact of concurrency value" do
+      before(:each) do
+        allow(Chef::Knife::WinrmSession).to receive(:new).and_return(session)
+        @winrms = Chef::Knife::Winrm.new(['-C', '3', '-m', 'localhost', '-x', 'testuser', '-P', 'testpassword', '--winrm-authentication-protocol', 'basic', 'echo helloworld'])
+      end
+      it "create sessions to equal number of concurrent value if concurrent option is given" do
```
Nimesh-Msys: As per the description, can we have a number check in expectation ?
Also in the upcoming test cases.

lib/chef/knife/winrm_knife_base.rb
Nimesh-Msys: My bad, `queue` is an object of class [`Queue`](https://ruby-doc.org/core-2.5.0/Queue.html). This is not an array.

spec/unit/knife/winrm_spec.rb
```diff
@@ -443,6 +443,22 @@
...
+        allow(Chef::Knife::WinrmSession).to receive(:new).and_return(session)
+        @winrms = Chef::Knife::Winrm.new(['-C', '3', '-m', 'localhost', '-x', 'testuser', '-P', 'testpassword', '--winrm-authentication-protocol', 'basic', 'echo helloworld'])
+      end
+      it "connect sessions if concurrent option is given" do
+        @winrms.config[:concurrency] = 1
+        allow(@winrms).to receive(:relay_winrm_command).and_return(true)
+        expect { @winrms.run_with_pretty_exceptions }.not_to raise_error
+      end
+      it "connect sessions if concurrent option is not given" do
+        allow(@winrm).to receive(:relay_winrm_command).and_return(true)
```
marcparadise: We've set -C 3 in the before block, so wouldn't it still be given here?  We might need to `@winrms.config.delete[:concurrency]`.

spec/unit/knife/winrm_spec.rb
```diff
@@ -443,6 +443,22 @@
       expect { @winrm.run_with_pretty_exceptions }.to raise_error(SystemExit)
     end
 
+    context "impact of concurrency value" do
```
marcparadise: I'd like to see these tests verifying that the requested concurrency limits are respected; or that no concurrency limit is put in place when it's not specified. That is something like: 

``` 
context "impact of concurrency value" do 
  context "when concurrency limit is not set" do 
    it "spawns a number of connection threads equal to the number of target nodes" do 
    end 
  end
  context "when concurrency limit is set" do 
     it "starts only the required number of threads when there are fewer targets than threads" do 
    end  
     it "starts only the requested number of threads when there are as many targets as threads" do 
    end 
     it "starts only the requested number of threads when there are more targets then threads" do 
    end 
   end 
end
```

That first test under when "concurrency limit is set" will fail if implemented as written now, because we start C threads even if number of targets < C. That seems like overhead we could avoid.

tas50:
Can this get rebased the comments addressed please?

lib/chef/knife/winrm_shared_options.rb
```diff
@@ -28,24 +28,23 @@ module WinrmSharedOptions
...
 
           option :concurrency,
-            :short => "-C NUM",
-            :long => "--concurrency NUM",
-            :description => "The number of allowed concurrent connections",
-            :default => 1,
-            :proc => lambda { |o| o.to_i }
+            short: "-C NUM",
+            long: "--concurrency NUM",
+            description: "The number of allowed concurrent connections",
```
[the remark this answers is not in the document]
NAshwini: Whether you want me to add this as a description ?

NAshwini:
@marcparadise and @tas50 Sorry for late response. I have refactor the code as per review comment and rebase it.

Thanks.

lib/chef/knife/winrm_shared_options.rb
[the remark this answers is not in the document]
marcparadise: Yes, to the description.

spec/unit/knife/winrm_spec.rb
```diff
@@ -443,6 +443,35 @@
...
+      before(:each) do
+        allow(Chef::Knife::WinrmSession).to receive(:new).and_return(session)
+        @winrms = Chef::Knife::Winrm.new(['-C', '3', '-m', 'localhost somewhere knowhost', '-x', 'testuser', '-P', 'testpassword', '--winrm-authentication-protocol', 'basic', 'echo helloworld'])
+      end
+      context "when concurrency limit is set" do
```
marcparadise: The shape of these is better, but I don't see that we're verifying the number of threads started in each case.

lib/chef/knife/winrm_knife_base.rb
```diff
@@ -122,11 +121,15 @@ def relay_winrm_command(command)
 
             queue = Queue.new
             @winrm_sessions.each { |s| queue << s }
+            con_limit = config[:concurrency]
+            num_sessions =  locate_config_value(:concurrency)
+            num_targets =  @winrm_sessions.length
+            num_sessions  = (con_limit.nil? || con_limit == 0) ? num_targets : [num_sessions, num_targets].min
```
marcparadise: A little confused by this - doesn't `locate_config_value` also search in `config`?

NAshwini: Actually it is but setting con_limit as hard code value might be not a good idea, So set it as `config[:concurrency]`

marcparadise: What I mean is, is there a scenario in which `locate_config_value` will fail to return the value included in `config[:concurrency]`?

NAshwini: Not require to set con_limit as `config[:concurrency]` will be taken care in def locate_config_value,
made changes as required.

spec/unit/knife/winrm_spec.rb
NAshwini: @mp I have added the test case as below, But as per your point we cannot check impact of concurreny with number of threads running as it does the actually checking of host connection. which returns to fail due to dummy node. and mocking it will not give result of number of threads.

spec/unit/knife/winrm_spec.rb
```diff
@@ -491,4 +491,65 @@
...
+        '--winrm-authentication-protocol', protocol,
+        'echo helloworld'
+      ]
+    end
+    let(:winrm_connection) { Dummy::Connection.new }
+
+    subject { Chef::Knife::Winrm.new(knife_args) }
+
+    it "concurrency not set" do
+      expect(subject).to receive(:run_command_in_thread).thrice
```
marcparadise: I think there's value in verifying that we've called run_command_in_thread three times, but we also want to verify the number of threads spawned in these cases. Could we do that via something like `expect(Thread).to receive(:new).exactly(n).times.and_call_original` ? That would validate that we're running the command for all hosts, and that the we are not spawning more (or fewer) threads than we should be in each case.

NAshwini: Yes I got your point, But when we tried to check number of spawning threads, it actually goes to make connection for all host under `-m` i.e `localhost knownhost somehost` and it fails to connect these dummy host and return a error. and mocking this will not help to get no of spawning thread.

marcparadise: I think we could get around that by continuing to check for receive `receive(:run_command_in_thread).thrice` as well; the thread would get spawned, but `run_command_in_thread` would be prevented from execution by the condition.

NAshwini: Thank you @marcparadise, Added test cases of `Threads` and rebase it.

PR #510, opened 2021-03-12
Fix test for ruby-3.0

lamont-granquist:
This just introduces something we can easier stub since $? is
apparently now considered frozen and rspec cannot stub it.

### DIFF
--- a/lib/chef/knife/windows_listener_create.rb
+++ b/lib/chef/knife/windows_listener_create.rb
@@ -1,5 +1,5 @@
 # Author:: Mukta Aphale (<mukta.aphale@clogeny.com>)
-# Copyright:: Copyright (c) 2014-2016 Chef Software, Inc.
+# Copyright:: Copyright (c) Chef Software Inc.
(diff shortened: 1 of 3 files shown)


lib/chef/knife/windows_listener_create.rb
```diff
@@ -60,6 +60,10 @@ def get_cert_passphrase
         passphrase.strip
       end
 
+      def exitstatus
+        $?
```
marcparadise, line 64: This should return `$?.exitstatus`, or our check of `if exitstatus` below will eval truthy when it shouldn't.

marcparadise (changes requested):
This comment is because github requires me to make it in order to request changes.  The real comment is inline.

lib/chef/knife/windows_listener_create.rb
lamont-granquist: ah yes, that's what i meant to do.  also why i don't particularly like unit tests.

marcparadise (approved):

PR #512, opened 2021-04-29
Support external testing

lamont-granquist:
The chef/chef repo needs to inject itself via ENV vars into the
Gemfile of this repo

(We have been uselessly testing knife-windows against the last release of chef/chef on every PR
in that repo instead of testing against the submitted code)


lamont-granquist:
This is busted, but that's kind of the point right now.  Now it should show up on chef/chef as well with every PR there.

PR #525, opened 2024-06-19
Added the ruby 3.3 pipeline

nikhil2611:
## Description
Added the ruby 3.3 pipeline

## Related Issue
https://chefio.atlassian.net/browse/CHEF-12436

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Chore (non-breaking change that does not add functionality or fix an issue)

## Checklist:

- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have run the pre-merge tests locally and they pass.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] If `Gemfile.lock` has changed, I have used `--conservative` to do it and included the full output in the Description above.
- [ ] All new and existing tests passed.
- [ ] All commits have been signed-off for [the Developer Certificate of Origin](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).

### DIFF
--- a/.expeditor/verify.pipeline.yml
+++ b/.expeditor/verify.pipeline.yml
@@ -18,6 +18,13 @@ steps:
     executor:
       docker:
         image: ruby:3.1-buster
+- label: run-lint-and-specs-ruby-3.3
+  command:
+    - .expeditor/run_linux_tests.sh rake
+  expeditor:
+    executor:
+      docker:
+        image: ruby:3.3
 - label: run-lint-and-specs-windows
   command:
     - gem install chef-utils


knife-windows.gemspec
```diff
@@ -12,8 +12,8 @@ Gem::Specification.new do |s|
   s.summary     = %q{Plugin that adds functionality to Chef Infra's Knife CLI for configuring/interacting with nodes running Microsoft Windows}
   s.description = s.summary
 
-  s.required_ruby_version = ">= 3.1"
-  s.add_dependency "chef", ">= 18.2"
+  s.required_ruby_version = ">= 3.3"
```
Stromweld, line 15: I wouldn't update this to 3.3 as it then wont be able to support chef-18. This does not control what version of ruby this works with other than setting a minimum version of ruby. Unless there is a breaking change or we are bumping for limiting versions we support this shouldn't get bumped up then. The ruby version installed on the system determines which ruby version this is working with. Expeditor is where you'd define the ruby version to run and test this on. @tpowell-progress can help explain this more if needed.

nikhil2611: Yeah @Stromweld we wont we merging these PRs now, these pr's will be merged  after ruby 3.3 upgrade in chef-client and inspec.

Stromweld: in the end this isn't needed for that. >= 3.1 sets minimum ruby version for this gem, unless there is a reason to need to bump the minimum version such as a change that is only in a newer version it doesn't make sense to update the required_ruby_version as it'll create compatibility problems for chef 18 and older versions that are not running on ruby 3.3 but will still need to be supported.

nikhil2611: okay but ruby 3.1 will be going end of life by 2025-03-31, should be still keep that ?

Stromweld: I'd say yes since chef-client 18.x wont be EOL by then. Unless there are code changes that take advantage of newer ruby version changes that make it not backwards compatible with older ruby version then this value really doesn't need to change.

Gemfile
```diff
@@ -8,7 +8,7 @@ if ENV["GEMFILE_MOD"]
   puts "GEMFILE_MOD: #{ENV["GEMFILE_MOD"]}"
   instance_eval(ENV["GEMFILE_MOD"])
 else
-  gem "ohai", git: "https://github.com/chef/ohai", branch: "main"
+  gem "ohai", "~> 18.0"
```
tpowell-progress, line 11: This should be `18-stable`

knife-windows.gemspec
tpowell-progress: I can imagine that we will likely keep Ruby 3.1 in place for the foreseeable future, despite official EOL. I would in general be as conservative as possible with minimum Ruby requirements to avoid unnecessary headaches. Ruby 3.0 introduces some Ruby API changes that are sensible to require, but I would hesitate be too much higher than that. Considering this is updates going forward, Ruby 3.1 is understandable.

tpowell-progress (changes requested):
Ruby 3.1 and use `18-stable`

